Issue #11, opened 2025-07-18
Avoid GitHub API rate limiting in telemetry setup


jerop:
The action's telemetry setup script ([scripts/telemetry.js](https://github.com/google-github-actions/run-gemini-cli/blob/main/scripts/telemetry.js)) downloads the OpenTelemetry Collector binary from a GitHub release. This download process involves making requests to the GitHub API to get the latest release information.

Currently, these API requests are unauthenticated. Unauthenticated requests are subject to a strict rate limit of [60 requests per hour](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-unauthenticated-users) per IP address. In active environments, this limit can be easily exhausted, causing the run-gemini-cli action to fail with an HTTP 403 error during the "Setup Telemetry Collector" step.

By making an authenticated API request, the rate limit will be increased from 60 to [5,000 requests per hour](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-authenticated-users), which is sufficient for typical usage and prevents these failures.

PR #12, opened 2025-07-18
Fix: Avoid GitHub API rate limiting in telemetry setup

jerop:
The action's telemetry setup script downloads the OpenTelemetry Collector binary from a GitHub release. This download process involves making requests to the GitHub API to get the latest release information.

Currently, these API requests are unauthenticated and are subject to a strict rate limit of 60 requests per hour per IP address. This change updates the script to use the 'GITHUB_TOKEN' environment variable when making requests to the GitHub API. The script now checks for the presence of 'process.env.GITHUB_TOKEN' and adds the 'Authorization' header if it is available.

By making an authenticated API request, the rate limit is increased from 60 to 5,000 requests per hour, which is sufficient for typical usage and prevents these failures. This change is backward-compatible and will fall back to unauthenticated requests if the 'GITHUB_TOKEN' is not available.

We can optimize the OpenTelemetry Collector setup in future work, such as running it via a docker container.

Fixes https://github.com/google-github-actions/run-gemini-cli/issues/11.

### DIFF
--- a/scripts/telemetry_utils.js
+++ b/scripts/telemetry_utils.js
@@ -32,7 +32,14 @@ export function getJson(url) {
   );
   try {
     const headers = ['-H "User-Agent: gemini-cli-dev-script"'];
-    if (process.env.GITHUB_TOKEN && url.startsWith('https://api.github.com')) {
+    const parsedUrl = new URL(url);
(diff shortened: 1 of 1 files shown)


jerop:
Closing this in favor of https://github.com/google-github-actions/run-gemini-cli/pull/15

PR #13, opened 2025-07-18
feat: Add generic workflow and example (gemini-cli)

galz10:
* Generic Prompting: Unlike the specialized /review or /triage slash commands, the `@gemini-cli` command is designed for general-purpose interaction. Users can ask any question or give any instruction, and the CLI will process it.
   * `examples/gemini-cli.yml`: A new workflow that triggers on PR comments containing `@gemini-cli`. It checks out the PR branch, runs the Gemini CLI with the comment body as the user request, and pushes any resulting code changes back to the branch.
   * `.gitignore`: Updated to ignore temporary gha-creds-*.json files generated by the create-github-app-token action.


.github/workflows/gemini-cli.yml
```diff
@@ -0,0 +1,255 @@
...
+  pull_request_review:
+    types:
+      - 'submitted'
+  issue_comment:
+    types:
+      - 'created'
+  issues:
+    types:
+      - 'opened'
+      - 'assigned'
```
jerop: what's the use case for issue being opened or assigned triggering this generic workflow?

.github/workflows/gemini-cli.yml
```diff
@@ -0,0 +1,255 @@
+name: '💬 Gemini CLI'
+
+on:
+  pull_request:
+    types:
+      - 'opened'
```
jerop: what's the use case for PR being opened triggering this generic workflow?

galdawave:
> I was expecting this workflow to only respond to comments coming from:
> 
> * [issue_comment](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#issue_comment)
> * [pull_request_comment](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_review_comment)
> * [pull_request_review_comment](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_review_comment)
> 
> And maybe from [discussion_comment](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#discussion_comment) in the future

I figured users might be able to generate FR plans immediately for example `Create a plan to add xyz` but i'll remove it for now the inital idea is to only support comments so might be a safer approach for initial release

.github/workflows/gemini-cli.yml
galdawave: Removed.

.github/workflows/gemini-cli.yml
galdawave: Removed.

PR #32, opened 2025-07-23
Add a max session turns for workflow

aliciatang07:
configure maxsessionturn


aliciatang07:
added as readme as discussed

PR #38, opened 2025-07-25
Adding GOOGLE_API_KEY as a Vertex Auth Option

HassanJasim:
You are allowed to authenticate with GOOGLE_API_KEY or GOOGLE_CLOUD_PROJECT & GOOGLE_CLOUD_LOCATION when using Vertex. This PR adds support for GOOGLE_API_KEY


jerop:
@gemini-cli please create a commit to address the feedback: while the action.yml correctly passes the GOOGLE_API_KEY, the action's documentation (likely in README.md) should be updated to inform users about this new authentication option and how to configure it.

jerop:
@HassanJasim is this ready for review? if so, please change it from "draft" to "ready"

PR #41, opened 2025-07-27
Add token exchange endpoint

squee1945:
Add terraform to create Google Cloud infrastructure to host a token exchange endpoint.

Add a workflow to build and deploy the token exchange endpoint, based on `github.com/abcxyz/github-token-minter` to Cloud Run.

To use, we will need to create a Google Cloud project to host the infrastructure.
We will also need to select a repository to host the CI pipeline (presumably `google-github-actions/run-gemini-cli`).

run-gemini-cli onboarding will need to be updated to create an authorization file at `.gemini/run-gemini-cli-auth.yaml` in each customer repository (see `./token-exchange/README.md` for info). The contents of this file are suggested in the README.md, but these are flexible and we may want a different setup.

The composite action will need to be updated to perform the token exchange flow, if appropriate (example in `./token-exchange/README.md`).

cc @jerop

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,9 +65,9 @@ Thumbs.db
 
 # GitHub App credentials
 gha-creds-*.json
-token-exchange/tf/.terraform.lock.hcl
 token-exchange/tf/terraform.tfstate
 token-exchange/tf/terraform.tfstate.backup
-token-exchange/tf/.terraform/providers/registry.terraform.io/hashicorp/google/6.45.0/darwin_arm64/LICENSE.txt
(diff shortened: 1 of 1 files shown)


sethvargo:
Hi @squee1945 - all our infrastructure exists at github.com/google-github-actions/test-infra, and there should already be a WIF pool configured.

squee1945:
> Hi @squee1945 - all our infrastructure exists at github.com/google-github-actions/test-infra, and there should already be a WIF pool configured.

Thanks @sethvargo - the WIF pools in this CL are in support of a prototype/alpha production service. Would `test-infra` support this use case?

sethvargo:
Hey @squee1945 - if this is going to be a public endpoint that we (Google) manage, then I don't think this repo (or test-infra) is the right place. Do you also intend to deploy Cloud Run services in this Terraform?

PR #42, opened 2025-07-28
feat: Add Issue Deduplication Workflow

gsquared94:
This pull request introduces an automated issue deduplication system workflow. This system uses an MCP server to generate and compare semantic embeddings of GitHub issues, allowing it to identify and report potential duplicates.

 ### Core Feature: Issue Deduplication

   * MCP Server (`scripts/issue-deduplication`):
       * A new FastMCP server is introduced, which connects to an AlloyDB database to store and query issue embeddings.
       * Provides two main tools:
           * refresh: Updates the embeddings for all open issues in a repository.
           * duplicates: Finds issues that are semantically similar to a given issue.
       * Includes a Dockerfile for easy containerization and a setup_alloydb.sh script to automate the provisioning of the required Google Cloud resources.

   * GitHub Actions Workflows:
       * Real-time Duplicate Detection: The gemini-issue-automated-triage.yml workflow now includes a deduplicate-issues job. This job is triggered when a new issue is opened or when a user comments `@gemini-cli /deduplicate`. It calls the MCP server to find duplicates and posts a comment on the issue with the results, and add the `duplicate` label.
       * Scheduled Embedding Refresh: The gemini-issue-scheduled-triage.yml workflow now includes a refresh-embeddings job that runs on a schedule to keep the AlloyDB database up-to-date with the latest issue embeddings.

 ### Future work

   * Update the triggered workflow to allow the bot to automatically close an issue as duplicate (for high enough confidence).
   * Update the scheduled workflow to perform automatic deduplication across all open issues.

### DIFF
--- a/.github/workflows/gemini-issue-automated-triage.yml
+++ b/.github/workflows/gemini-issue-automated-triage.yml
@@ -138,18 +138,20 @@ jobs:
           app-id: '${{ vars.APP_ID }}'
           private-key: '${{ secrets.APP_PRIVATE_KEY }}'
 
-      - id: 'auth'
+      - name: 'Authenticate to Google Cloud'
+        id: 'auth'
         uses: 'google-github-actions/auth@140bb5113ffb6b65a7e9b937a81fa96cf5064462' # ratchet:google-github-actions/auth@v2
         with:
+          token_format: 'access_token'
           workload_identity_provider: '${{ vars.GCP_WIF_PROVIDER }}'
           service_account: '${{ vars.SERVICE_ACCOUNT_EMAIL }}'
 
-      - name: 'Set up Cloud SDK'
-        uses: 'google-github-actions/setup-gcloud@6189d56e4096ee891640bb02ac264be376592d6a' # ratchet:google-github-actions/setup-gcloud@v2
-
-      - name: 'Configure Docker for Artifact Registry'
-        run: |-
-          gcloud auth configure-docker northamerica-northeast1-docker.pkg.dev
+      - name: 'Login to GAR'
+        uses: 'docker/login-action@v3'
+        with:
+          registry: 'northamerica-northeast1-docker.pkg.dev'
+          username: 'oauth2accesstoken'
+          password: '${{ steps.auth.outputs.access_token }}'
 
       - name: 'Download AlloyDB Auth Proxy'
         run: |
--- a/.github/workflows/gemini-issue-automated-triage.yml
+++ b/.github/workflows/gemini-issue-automated-triage.yml
@@ -173,8 +173,6 @@ jobs:
           GITHUB_TOKEN: '${{ steps.generate_token.outputs.token || secrets.GITHUB_TOKEN }}'
           ISSUE_NUMBER: '${{ github.event.issue.number }}'
           REPOSITORY: '${{ github.repository }}'
-          REPO_OWNER: '${{ github.repository_owner }}'
-          REPO_NAME: '${{ github.event.repository.name }}'
           GEMINI_CLI_VERSION: '${{ vars.GEMINI_CLI_VERSION }}'
           OTLP_GOOGLE_CLOUD_PROJECT: '${{ vars.OTLP_GOOGLE_CLOUD_PROJECT }}'
           GCP_WIF_PROVIDER: '${{ vars.GCP_WIF_PROVIDER }}'
@@ -243,8 +241,8 @@ jobs:
             ## Steps
 
             1.  **Find Potential Duplicates:**
-                - The repo_owner/repo_name is ${{ github.repository }} and the issue number is ${{ github.event.issue.number }}.
-                - Use the `duplicates` tool with the `repo_owner`, `repo_name`, and `issue_number` to find potential duplicates for the current issue. Do not use the `threshold` parameter.
+                - The repository is ${{ github.repository }} and the issue number is ${{ github.event.issue.number }}.
+                - Use the `duplicates` tool with the `repo` and `issue_number` to find potential duplicates for the current issue. Do not use the `threshold` parameter.
                 - If no duplicates are found, you are done.
 
             2.  **Refine Duplicates List (if necessary):**
--- a/.github/workflows/gemini-issue-automated-triage.yml
+++ b/.github/workflows/gemini-issue-automated-triage.yml
@@ -116,13 +116,14 @@ jobs:
 
   deduplicate-issues:
     if: >
-      github.event_name == 'issues' ||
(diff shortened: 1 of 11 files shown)


.github/workflows/gemini-issue-automated-triage.yml
```diff
@@ -113,3 +113,186 @@ jobs:
...
+                    "GEMINI_API_KEY",
+                    "-e",
+                    "DB_USER",
+                    "-e",
+                    "DB_PASS",
+                    "-e",
+                    "INSTANCE_HOST",
+                    "-e",
+                    "INSTANCE_PORT",
+                    "northamerica-northeast1-docker.pkg.dev/quacktastic-waffle/run-gemini-cli/issues-dedup-mcp-server:latest"
```
sethvargo, line 215: This doesn't seem very official?

.github/workflows/gemini-issue-automated-triage.yml
```diff
@@ -113,3 +113,186 @@ jobs:
...
+    runs-on: 'ubuntu-latest'
+
+    steps:
+      - name: 'Checkout repository'
+        uses: 'actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683' # ratchet:actions/checkout@v4
+
+      - name: 'Generate GitHub App Token'
+        id: 'generate_token'
+        if: |-
+          ${{ vars.APP_ID }}
```
sethvargo:
```suggestion
          ${{ vars.APP_ID != "" }}
```

or some kind of specific check.

Throughout other conditionals too.

.github/workflows/gemini-issue-automated-triage.yml
```diff
@@ -113,3 +113,186 @@ jobs:
...
+        uses: 'google-github-actions/setup-gcloud@6189d56e4096ee891640bb02ac264be376592d6a' # ratchet:google-github-actions/setup-gcloud@v2
+
+      - name: 'Configure Docker for Artifact Registry'
+        run: |-
+          gcloud auth configure-docker northamerica-northeast1-docker.pkg.dev
+
+      - name: 'Download AlloyDB Auth Proxy'
+        run: |
+          curl -o alloydb-auth-proxy https://storage.googleapis.com/alloydb-auth-proxy/v1.13.4/alloydb-auth-proxy.linux.amd64
+          chmod +x alloydb-auth-proxy
```
sethvargo, line 161: AlloyDB seems like overkill for this use case? We're just loading some issues. Honestly, Firestore or Datastore might be a better option unless you need the SQL-ness.

.github/workflows/gemini-issue-automated-triage.yml
```diff
@@ -113,3 +113,186 @@ jobs:
...
+        with:
+          workload_identity_provider: '${{ vars.GCP_WIF_PROVIDER }}'
+          service_account: '${{ vars.SERVICE_ACCOUNT_EMAIL }}'
+
+      - name: 'Set up Cloud SDK'
+        uses: 'google-github-actions/setup-gcloud@6189d56e4096ee891640bb02ac264be376592d6a' # ratchet:google-github-actions/setup-gcloud@v2
+
+      - name: 'Configure Docker for Artifact Registry'
+        run: |-
+          gcloud auth configure-docker northamerica-northeast1-docker.pkg.dev
```
sethvargo: Is this just for pulling the container? Sorry I'm not sure why gcloud is needed here. You can pull the cred from the auth directly into `docker pull` to save a ton of time: https://github.com/docker/login-action?tab=readme-ov-file#workload-identity-federation-1

.github/workflows/gemini-issue-automated-triage.yml
```diff
@@ -113,3 +113,186 @@ jobs:
...
+                  "command": "docker",
+                  "args": [
+                    "run",
+                    "-i",
+                    "--rm",
+                    "--network=host",
+                    "-e",
+                    "GITHUB_TOKEN",
+                    "-e",
+                    "GEMINI_API_KEY",
```
sethvargo, line 206: Why does the container need the Gemini API Key? What interactions is it doing with Gemini?

gsquared94: It is used to generate the embeddings. See https://github.com/google-github-actions/run-gemini-cli/pull/42/files#diff-12fc83002d147443fedf7c3e4b7b2331a049a59174fbf3d5636031ba5e8d4f89R194-R197

.github/workflows/gemini-issue-automated-triage.yml
gsquared94: We have a lot of resource for this repo being provisioned in the `quacktastic-waffle` project. The intention is for external users to build their own image from source (`scripts/issue-deduplication/Dockerfile`) and not use this image (it is not publicly accessible). WDYT?

.github/workflows/gemini-issue-automated-triage.yml
gsquared94: Done

.github/workflows/gemini-issue-automated-triage.yml
gsquared94: Done.

.github/workflows/gemini-issue-automated-triage.yml
gsquared94: It was easier to get working with alloyDB. Let me investigate firestore and it's vector search capabilities.

.github/workflows/gemini-issue-automated-triage.yml
sethvargo: Hmm - asking developers to build containers and run databases seems like a lot of overhead relative to the feature functionality imo. It can be a follow-up, but we should think about how to reduce the overhead.

For example, we could vend Terraform configurations for a one-step bootstrapping. Or we could decouple entirely from running infrastructure and just have an MCP server that exposes a tool for getting all the GitHub issues in a given repo. I looked and even the most popular repos like homebrew only have 20k issues (and most are PRs). The largest I found was vscode with 260k issues. All of that should fit in memory, but most projects are probably significantly smaller and could paginate the GitHub API and do an in-memory search.

gsquared94: Sounds good. I'll setup the image on this repo's container registry instead with public access and fast follow with another PR to use that here. I'll also fast follow with a PR to add both firestore and in-memory database options. 

IMO the cost and overhead of inference for recomputing the embeddings makes a persistent storage useful. I'm also testing approaches for dimensionality reduction to fit in the embedding model token limits better which will also increase the cost of embedding.

PR #46, opened 2025-07-29
docs(readme): clarify inputs and add links

leehagoodjames:
Clarify how the inputs are passed to the Gemini CLI and add links to docs that Gemini CLI uses to define these terms.

Updates were made to the `actions.yml` file, and then the readme was updated by running `npm run docs`

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -19,13 +19,14 @@ description: |-
 
 inputs:
   prompt:
-    description: 'A specific prompt to guide Gemini.'
+    description: |
+      A string passed to the Gemini CLI's [`--prompt` argument](https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/configuration.md#command-line-arguments).
     required: false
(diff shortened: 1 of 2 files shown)


action.yml
```diff
@@ -19,14 +19,16 @@ description: |-
...
     required: false
     default: 'You are a helpful assistant.'
   settings:
     description: |
-      A JSON string to configure the Gemini CLI. This will be written to
-      .gemini/settings.json.
+      A JSON string written to `.gemini/settings.json` to configure the CLI's _project_ settings.
+      For more details, see the documentation on [settings files](https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/configuration.md#settings-files).
     required: false
+    default: None
```
sethvargo: You can probably omit this line

leehagoodjames: Updated to omit this line

PR #57, opened 2025-07-31
fix(workflows): simplify manual trigger for scheduled triage

jerop:
This commit updates the scheduled issue triage workflow to simplify manual runs and align it with the source workflow.

The `issue_number` input for the `workflow_dispatch` trigger has been removed. The workflow now automatically identifies all untriaged issues, so a manual issue selection is no longer needed. This allows for manually triggering a triage run for all relevant issues at once.

Additionally, the step that posts a failure comment was removed to avoid spamming issues if the scheduled triage fails.

### DIFF
--- a/.github/workflows/gemini-issue-scheduled-triage.yml
+++ b/.github/workflows/gemini-issue-scheduled-triage.yml
@@ -4,11 +4,6 @@ on:
   schedule:
     - cron: '0 * * * *' # Runs every hour
   workflow_dispatch:
-    inputs:
-      issue_number:
-        description: 'issue number to triage'
-        required: true
-        type: 'number'
 
 concurrency:
   group: '${{ github.workflow }}'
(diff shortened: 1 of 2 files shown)


workflows/issue-triage/gemini-issue-scheduled-triage.yml
```diff
@@ -26,13 +26,13 @@ jobs:
 
     steps:
       - name: 'Checkout repository'
-        uses: 'actions/checkout@v4'
+        uses: 'actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683' # ratchet:actions/checkout@v4
```
leehagoodjames, line 29: What is the intent of fixing to a specific version?

.github/workflows/gemini-issue-scheduled-triage.yml
```diff
@@ -127,17 +122,3 @@ jobs:
             - Do not add comments
             - Triage each issue independently
             - Reference all shell variables as "${VAR}" (with quotes and braces)
-
-      - name: 'Post Issue Triage Failure Comment'
```
JeromeJu, line 131: Do we not need failure feedback comment when it's scheduled triaging?

jerop: no, because this could spam issues -- let's only post a comment when it's automated workflows -- also scheduled workflows get retried, if transient issues

workflows/issue-triage/gemini-issue-scheduled-triage.yml
jerop: This is a security best practice - https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions 
```
Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload. When selecting a SHA, you should verify it is from the action's repository and not a repository fork.
```

PR #59, opened 2025-07-31
feat: add failure notification to pr review workflow

jerop:
This adds a new step to the PR review workflow that posts a comment to the PR if the Gemini CLI review fails. This provides better visibility into workflow failures.

### DIFF
--- a/workflows/pr-review/gemini-pr-review.yml
+++ b/workflows/pr-review/gemini-pr-review.yml
@@ -66,15 +66,13 @@ jobs:
 
     steps:
       - name: 'Checkout PR code'
-        uses: 'actions/checkout@v4'
-        with:
-          fetch-depth: 0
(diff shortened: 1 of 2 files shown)


workflows/pr-review/gemini-pr-review.yml
```diff
@@ -302,3 +301,17 @@ jobs:
             - List any questions you have about the implementation
             - Clarifications needed from the author
             '''
+
+      - name: 'Post PR review failure comment'
```
leehagoodjames, line 305: Why do we want to add this for the PR Review workflow and not the Issue triage workflow?

jerop: We already have this for automated issue triage (just removed it from the bulk scheduled issue triage, not the automated one upon issue creation)

jerop: https://github.com/google-github-actions/run-gemini-cli/tree/main/workflows/issue-triage --> there are two types of issue triage workflows -- automated and scheduled

PR #61, opened 2025-07-31
feat: update workflow timeouts and document changes

jerop:
This commit introduces two main changes:

1. The default timeouts for the gemini-cli, gemini-issue-scheduled-triage, and gemini-pr-review workflows have been updated to more reasonable values to prevent excessive resource usage. The gemini-issue-automated-triage had a reasonable timeout already.

2. The workflows/README.md has been updated to include instructions on how users can customize the timeout-minutes for each workflow.

These changes improve the safety and usability of the provided example workflows.

### DIFF
--- a/.github/workflows/gemini-cli.yml
+++ b/.github/workflows/gemini-cli.yml
@@ -83,7 +83,7 @@ jobs:
           )
         )
       )
-    timeout-minutes: 15
+    timeout-minutes: 10
     runs-on: 'ubuntu-latest'
 
     steps:
--- a/workflows/README.md
+++ b/workflows/README.md
@@ -8,6 +8,22 @@ This directory contains a collection of example workflows that demonstrate how t
 *   **[Pull Request Review](./pr-review)**: Automatically review pull requests using Gemini. This workflow can be triggered by pull request events and provides a comprehensive review of the changes.
 *   **[Gemini CLI](./gemini-cli)**: A general-purpose, conversational AI assistant that can be invoked within pull requests and issues to perform a wide range of tasks.
 
+## Configuration
+
+### Workflows
+
+#### Timeouts
+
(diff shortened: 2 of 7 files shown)


.github/workflows/gemini-cli.yml
```diff
@@ -83,7 +83,7 @@ jobs:
           )
         )
       )
-    timeout-minutes: 15
+    timeout-minutes: 10
```
leehagoodjames, line 86: I was curious about these, do they cancel inflight requests to the Gemini API when a timeout occurs, or just shutdown the action and prevent future requests?

workflows/README.md
```diff
@@ -8,6 +8,18 @@ This directory contains a collection of example workflows that demonstrate how t
 *   **[Pull Request Review](./pr-review)**: Automatically review pull requests using Gemini. This workflow can be triggered by pull request events and provides a comprehensive review of the changes.
 *   **[Gemini CLI](./gemini-cli)**: A general-purpose, conversational AI assistant that can be invoked within pull requests and issues to perform a wide range of tasks.
 
+## Customizing Timeouts
+
+Each workflow includes a `timeout-minutes` property that you can customize to meet your needs. This property determines the maximum amount of time that a job can run before it is automatically canceled.
+
+To change the timeout for a workflow, open the corresponding YAML file and modify the value of the `timeout-minutes` property. For example, to change the timeout for the Pull Request Review workflow to 10 minutes, you would make the following change:
```
leehagoodjames, line 19: nit: May want to link to [job.step](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepstimeout-minutes) or [job](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idtimeout-minutes) timeout-minutes

.github/workflows/gemini-cli.yml
jerop: They shut down the job and project further execution

workflows/README.md
jerop: good idea, let me add this!

.github/workflows/gemini-cli.yml
leehagoodjames: Also - for open-ended tasks, I have used agents that take ~45 minutes, and I prefer a ~1hr timeout. 

For complex tasks, it may take a long time - so canceling the request causes the user to still pay but not get any value

jerop: Got some guidance from the quality workstream to make this smaller so that we can control for looping or other issues, then we can give users docs to set a higher timeout if needed

PR #70, opened 2025-07-31
fix: pr-review workflow tool inclusion list for mcp server

gsquared94:
fix: pr-review workflow tool inclusion list for mcp server

### DIFF
--- a/.github/workflows/gemini-pr-review.yml
+++ b/.github/workflows/gemini-pr-review.yml
@@ -175,9 +175,9 @@ jobs:
                     "ghcr.io/github/github-mcp-server"
                   ],
                   "includeTools": [
-                    "mcp__github__create_pending_pull_request_review",
-                    "mcp__github__add_comment_to_pending_review",
-                    "mcp__github__submit_pending_pull_request_review"
+                    "create_pending_pull_request_review",
+                    "add_comment_to_pending_review",
+                    "submit_pending_pull_request_review"
                   ],
                   "env": {
                     "GITHUB_PERSONAL_ACCESS_TOKEN": "${GITHUB_TOKEN}"
--- a/workflows/pr-review/gemini-pr-review.yml
+++ b/workflows/pr-review/gemini-pr-review.yml
@@ -175,9 +175,9 @@ jobs:
                     "ghcr.io/github/github-mcp-server"
                   ],
                   "includeTools": [
-                    "mcp__github__create_pending_pull_request_review",
-                    "mcp__github__add_comment_to_pending_review",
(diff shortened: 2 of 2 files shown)


gsquared94:
> Looks great, can you include test-results (such as running the workflow) to make sure that the workflow can still call these tools?

This is the run result:
https://github.com/gsquared94/run-gemini-cli/actions/runs/16661826020/job/47160495997

It seems to be calling into `google-github-actions/run-gemini-cli` instead of my fork, and failing to update it. But the tools seem to be working correctly.

jerop:
Looks like it failed to post a comment on this workflow: https://github.com/google-github-actions/run-gemini-cli/actions/runs/16661733176/job/47160197017#step:6:1

Wonder if we need to define it both in 'includeTools' and 'coreTools'? If so, we should improve this in Gemini CLI in future work

leehagoodjames:
> Looks like it failed to post a comment on this workflow: https://github.com/google-github-actions/run-gemini-cli/actions/runs/16661733176/job/47160197017#step:6:1
> 
> Wonder if we need to define it both in 'includeTools' and 'coreTools'? If so, we should improve this in Gemini CLI in future work

This is weird, since that run used `main`. It seems it loaded the code changes in this PR and thought it couldn't use certain tools, but it actually could since it was running the config in `main`.

jerop:
> > Looks like it failed to post a comment on this workflow: https://github.com/google-github-actions/run-gemini-cli/actions/runs/16661733176/job/47160197017#step:6:1
> > Wonder if we need to define it both in 'includeTools' and 'coreTools'? If so, we should improve this in Gemini CLI in future work
> 
> This is weird, since that run used `main`. It seems it loaded the code changes in this PR and thought it couldn't use certain tools, but it actually could since it was running the config in `main`.

@leehagoodjames don't think it's reading from main - it's using the branch - https://github.com/google-github-actions/run-gemini-cli/actions/runs/16661733176/workflow

leehagoodjames:
> > > Looks like it failed to post a comment on this workflow: https://github.com/google-github-actions/run-gemini-cli/actions/runs/16661733176/job/47160197017#step:6:1
> > > Wonder if we need to define it both in 'includeTools' and 'coreTools'? If so, we should improve this in Gemini CLI in future work
> > 
> > 
> > This is weird, since that run used `main`. It seems it loaded the code changes in this PR and thought it couldn't use certain tools, but it actually could since it was running the config in `main`.
> 
> @leehagoodjames don't think it's reading from main - it's using the branch - https://github.com/google-github-actions/run-gemini-cli/actions/runs/16661733176/workflow

Ya you are right, it looks like it was pulling in these changes

PR #75, opened 2025-08-01
PR Review Prompt Changes

leehagoodjames:
### Description
Update `.github/workflows/gemini-pr-review.yml` to try and make the prompt yield a more consistent result with a comments that are inline, prefer suggestions, have individual priorities, and a more concise summary. 

### Changes
- Remove references to the `write_file` tool, since it is no longer needed for posting the review
- Combine the comment syntaxes for `prioritization` and for `content` to a unified syntax
- Move `Review Areas` above the detailed approach describing tool use
- Indention changes for consistency
- Include the `summary` section with the other instructions, since the `summary` comment is made with the `mcp__github__submit_pending_pull_request_review` tool call
- Copy changes to `workflows/pr-review/gemini-pr-review.yml` 

### Testing
Invoked the action multiple times. Example pr review:
image

### DIFF
--- a/workflows/pr-review/gemini-pr-review.yml
+++ b/workflows/pr-review/gemini-pr-review.yml
@@ -142,7 +142,7 @@ jobs:
           } >> "${GITHUB_OUTPUT}"
 
       - name: 'Run Gemini PR Review'
-        uses: './'
+        uses: 'google-github-actions/run-gemini-cli@main'
         id: 'gemini_pr_review'
         env:
           GITHUB_TOKEN: '${{ steps.generate_token.outputs.token || secrets.GITHUB_TOKEN }}'


workflows/pr-review/gemini-pr-review.yml
```diff
@@ -142,7 +142,7 @@ jobs:
           } >> "${GITHUB_OUTPUT}"
 
       - name: 'Run Gemini PR Review'
-        uses: 'google-github-actions/run-gemini-cli@main'
+        uses: './'
```
jerop: please revert this, needs to point to the action

.github/workflows/gemini-pr-review.yml
```diff
@@ -272,37 +244,60 @@ jobs:
...
+            1. Creating a pending review: Use the mcp__github__create_pending_pull_request_review to create a Pending Pull Request Review.
+
+            2. Adding review comments:
+                2.1 Use the mcp__github__add_comment_to_pending_review to add comments to the Pending Pull Request Review. Inline comments are preffered whenever possible, so repeat this step, calling mcp__github__add_comment_to_pending_review, as needed. All comments about specific lines of code should use inline comments. It is preferred to use code suggestions when possible, which include a code block that is labeled "suggestion", which contains what the new code should be. All comments should also have a piority. They syntax is:
+                  Normal Comment Syntax:
+                  <COMMENT>
+                  {{PRIORITY}} {{COMMENT_TEXT}}
+                  </COMMENT>
+
+                  Inline Comment Syntax: (Preffered):
```
jerop:
```suggestion
                  Inline Comment Syntax: (Preferred):
```

workflows/pr-review/gemini-pr-review.yml
leehagoodjames: Undid the change for this `workflows/` file in 35845b5. Should this line be different in both files, or do we need to change the `.github/workflows/gemini-pr-review.yml` file too?

jerop: `.github/workflows/gemini-pr-review.yml` should be `uses: './'` because it's referring to itself -- this is the one line that's supposed to be different in the workflows

PR #77, opened 2025-08-01
Add Prerelease option for release

aliciatang07:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "run-gemini-cli",
-  "version": "0.1.0",
+  "version": "0.1.0-beta.0",
   "description": "This works with our versioning tools, this is NOT an NPM repo",
   "scripts": {
     "build": "echo \"No build required for composite action\"",


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "run-gemini-cli",
-  "version": "0.1.0",
+  "version": "0.1.0-beta.0",
```
sethvargo, line 3: Please keep this as "0.1.0" for now. When you trigger the release action, the automation will create a PR that re-compiles all the JS and updates this file. If you commit as-is, then when you release, it will create 0.1.0-beta.1.

aliciatang07: I think we want a release tag with beta right? if keeping the current version, it will not have a tag as beta version.

sethvargo: @aliciatang07 yes, but they will happen when you trigger the "draft-release" GitHub Action.

- Draft Release creates a PR that bumps the package.json and recompiles all the JS/TS
- Merging that PR does an actual release on GitHub (with the version information from the PR)

I just did it this morning on another repo, so you can see what I mean:

- I [ran the Draft Release workflow](https://github.com/google-github-actions/auth/actions/runs/16678021679), which created https://github.com/google-github-actions/auth/pull/503
- I merged https://github.com/google-github-actions/auth/pull/503, which [triggered the release](https://github.com/google-github-actions/auth/actions/runs/16678683092)

aliciatang07: Yeah I know if we didn't change the version, when running the draft release action, we will have a release with tag v0.1.1, is that what we want? I thought we want a tag with beta version. @jerop

aliciatang07: for the tag I mean this 
image

sethvargo: Sorry. Keep the change in "draft-release.yml" to allow the "prerelease" drop-down item. Revert the change to `package.json` though.

jerop: > I thought we want a tag with beta version

@aliciatang07 what i meant is that we want a tag named 'beta' in here: https://github.com/google-github-actions/run-gemini-cli/tags (much like the 'latest' label)

but the release itself uses semantic versioning with `MAJOR.MINOR.PATCH` format

PR #80, opened 2025-08-01
feat(workflows): Use GITHUB_TOKEN for checkout

galz10:
The actions/checkout action officially uses the GITHUB_TOKEN input to authenticate. This PR updates the following
  workflows to use the correct input name:

   - .github/workflows/gemini-cli.yml
   - workflows/gemini-cli/gemini-cli.yml

  This change ensures that the workflows are using the documented and supported method for authentication with the
  actions/checkout action.

### DIFF
--- a/.github/workflows/gemini-cli.yml
+++ b/.github/workflows/gemini-cli.yml
@@ -209,7 +209,7 @@ jobs:
           ${{  steps.get_context.outputs.is_pr == 'true' }}
         uses: 'actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683' # ratchet:actions/checkout@v4
         with:
-          token: '${{ steps.generate_token.outputs.token || secrets.GITHUB_TOKEN }}'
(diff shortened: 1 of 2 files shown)


.github/workflows/gemini-cli.yml
```diff
@@ -209,7 +209,7 @@ jobs:
           ${{  steps.get_context.outputs.is_pr == 'true' }}
         uses: 'actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683' # ratchet:actions/checkout@v4
         with:
-          token: '${{ steps.generate_token.outputs.token || secrets.GITHUB_TOKEN }}'
+          GITHUB_TOKEN: '${{ steps.generate_token.outputs.token || secrets.GITHUB_TOKEN }}'
```
jerop, line 212: is this supposed to be an env var?

don't see this as an input here: https://github.com/actions/checkout?tab=readme-ov-file#usage

PR #156, opened 2025-08-08
feat(telemetry): Specify gemini-SURFACE

leehagoodjames:
Fixes #155

PR #158, opened 2025-08-08
Revert "feat(telemetry): Specify gemini-SURFACE (#156)"

leehagoodjames:
This reverts commit 9da55331df5ce66d7b676a3a134e3abe13b7d00c.


jerop:
This is because we have data pipelines relying on this value

.github/workflows/gemini-pr-review.yml
```diff
@@ -150,7 +150,6 @@ jobs:
           CHANGED_FILES: '${{ steps.get_pr.outputs.changed_files || steps.get_pr_comment.outputs.changed_files }}'
           ADDITIONAL_INSTRUCTIONS: '${{ steps.get_pr.outputs.additional_instructions || steps.get_pr_comment.outputs.additional_instructions }}'
           REPOSITORY: '${{ github.repository }}'
-          SURFACE: 'GITHUB_ACTION'
```
jerop, line 153: this should be reset back to `GitHub`

jerop (changes requested):

.github/workflows/gemini-pr-review.yml
jerop: ah my bad, it's now relying on action.yml

PR #176, opened 2025-08-13
fix: output interruption after first response

ibezgin-bc:
Problem
The Gemini CLI command was stopping early when capturing output using direct shell command substitution ($(gemini ...)), likely due to buffering issues or incomplete output capture.

Solution
Replaced the direct command substitution with a more robust approach:

Temporary file capture: Uses mktemp to create a temporary file for capturing the full output
Explicit error handling: Added proper exit code checking to distinguish between success and failure
Complete output capture: Captures both stdout and stderr (2>&1) to ensure no output is lost
Proper cleanup: Temporary files are cleaned up after use

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -147,7 +147,16 @@ runs:
         fi
 
         # Run Gemini CLI with the provided prompt
-        GEMINI_RESPONSE=$(gemini --yolo --prompt "${PROMPT}")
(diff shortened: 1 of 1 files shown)


leehagoodjames:
@gemini-cli /review

sethvargo:
@leehagoodjames - I think this is actually why the CLI is hanging.

PR #171, opened 2025-08-13
feat: improve gemini issue triage workflow

jerop:
This commit improves the gemini issue triage workflow by:

- Passing the available labels as an environment variable to the Gemini CLI.
- Using the output of the Gemini CLI to apply labels to the issue.
- Adding a step to get all repository labels and pass them to the gemini-cli.
- Updating the prompt to classify the issue and output the labels in JSON format.
- Adding a step to apply the labels to the issue using the github-script.

image

PR #177, opened 2025-08-13
feat(workflows): reduce restrictions for invoking workflows for private repos

### LINKED ISSUE #163: author_association check in example workflows
Hi, 

I am testing out the gemini-pr-review workflow from the examples folder.

I noticed that it was being skipped when a pull request is created. The condition that was causing the skip is:
```
contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.pull_request.author_association)
```
Link to the line: https://github.com/google-github-actions/run-gemini-cli/blob/fa814daaa5974a0bd0630d9b90f72f29de664ea5/examples/workflows/pr-review/gemini-pr-review.yml#L45 

I am in a private repo and the value I get from `github.event.pull_request.author_association` is `CONTRIBUTOR` (even though the user opening the PR is a MEMBER).

Might be that this test can be modified to take into consideration the case I experienced. 

Additionally, I think in a private repo the test can be skipped. Would love to hear your thoughts about this.

leehagoodjames:
The workflows for https://github.com/google-github-actions/run-gemini-cli enable restrictions on who can invoke them, to prevent abuse scenarios by untrusted users. 

A better solution would enable reliably checking repository membership with `github.event.issue.author_association`, but this is not possible since `github.event.issue.author_association` can return `CONTRIBUTOR` even when the author is also a `MEMBER`. Given this, https://github.com/actions/github-script/issues/643 has been filed with GitHub to allow actions to more easily check for membership. 

This mitigation simplifies the experience for private repos

Fixes #163

### DIFF
--- a/.github/workflows/gemini-cli.yml
+++ b/.github/workflows/gemini-cli.yml
@@ -28,7 +28,9 @@ permissions:
 
 jobs:
   gemini-cli:
-    # This condition is complex to ensure we only run when explicitly invoked.
+    # This condition seeks to ensure the action is only run when it is triggered by a trusted user.
+    # For private repos, users who have access to the repo are considered trusted.
+    # For public repos, users who members, owners, or collaborators are considered trusted.
     if: |-
       github.event_name == 'workflow_dispatch' ||
       (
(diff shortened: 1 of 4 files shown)


leehagoodjames:
@gemini-cli /review

osis:
I've had to make this adjustment manually to use these workflows. Happy to see this going through! 🥳

sethvargo (approved):

sethvargo:
@leehagoodjames it looks like this is breaking on our repo now. https://github.com/google-github-actions/run-gemini-cli/pull/180 was skipped:
CleanShot 2025-08-13 at 16 20 23@2x

leehagoodjames:
> @leehagoodjames it looks like this is breaking on our repo now. #180 was skipped: CleanShot 2025-08-13 at 16 20 23@2x

Happy to revert if that is helpful, but we were seeing skips before this change because `github.event.issue.author_association` is returning `CONTRIBUTOR` (which you are) which is not supported in the list currently 

Example from https://github.com/google-github-actions/run-gemini-cli/pull/171
image

sethvargo:
Hmm okay. Weird

PR #180, opened 2025-08-13
chore(logging): only enable Gemini debug when DEBUG is set

sethvargo:
This is a follow-up to #178 which sets the Gemini CLI debug settings.


sethvargo:
I think #180 will fully capture error messages. When a user reports a bug, we can ask them to re-run the workflow with debug logging enabled (it's a checkbox is the UI):

CleanShot 2025-08-13 at 21 12 22@2x

That will also enable the debug logs for gemini.

PR #186, opened 2025-08-14
fix(output): separate the stdout and stderr logs

sethvargo:
This introduces a new output that holds only the error logs. Previously, the single error stream was being clobbered by both stderr and stdout writing to the same file. Also, you probably want the output values to be different.


sethvargo:
@gemini-cli /review

PR #203, opened 2025-08-20
Fix live streaming output whilst preserving stdout/stderr separation

iclectic:
Since #186, users wait until processes finish before seeing output. This uses tee with process substitution to stream live whilst maintaining separate stdout/stderr capture.

- Live output to console in real-time
- Stdout/stderr still captured separately
- Existing processing unchanged

Testing confirms both streaming and separation work correctly.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -161,7 +161,7 @@ runs:
 
         # Run Gemini CLI with the provided prompt
         FAILED=false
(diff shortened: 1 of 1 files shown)


sethvargo:
Hey @iclectic this is a good fix, thank you. However, we've gotten feedback that we shouldn't print output at all, since it's possible to trick the LLM into revealing information in the logs like secrets.

Issue #205, opened 2025-08-20
Support handling when maxSessionTurns causes early exit


leehagoodjames:
### TL;DR

`maxSessionTurns` is an optional parameter for gemini-cli [configuration](https://github.com/google-gemini/gemini-cli/blob/acedcfb8f7114b0a951cca33e39aaf7acee4da99/docs/cli/configuration.md#available-settings-in-settingsjson). 

When this limit is, gemini-cli exits. 

When this occurs for a GitHub action, the GitHub action should be failed and the appropriate cleanup should be done (ex: deleting comments of things being in-process, posting a failure comment, etc)

jerop:
Will be handled in https://github.com/google-github-actions/run-gemini-cli/issues/217

PR #206, opened 2025-08-20
feat(workflow): mark workloads as failed if they hit session limit

leehagoodjames:
Fixes: #205

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -189,6 +189,10 @@ runs:
         echo "${GEMINI_ERRORS}" >> "${GITHUB_OUTPUT}"
         echo "EOF" >> "${GITHUB_OUTPUT}"
 
+        if echo "${GEMINI_ERRORS}" | grep -q "Reached max session turns for this session"; then
+         FAILED=true
+        fi
+
         if [[ "${FAILED}" = true ]]; then
           LAST_LINE="$(echo "${GEMINI_ERRORS}" | tail -n1)"
           echo "::error title=Gemini CLI execution failed::${LAST_LINE}"


action.yml
```diff
@@ -189,6 +189,10 @@ runs:
         echo "${GEMINI_ERRORS}" >> "${GITHUB_OUTPUT}"
         echo "EOF" >> "${GITHUB_OUTPUT}"
 
+        if echo "${GEMINI_ERRORS}" | grep -q "Reached max session turns for this session"; then
```
jerop, line 192: if the error message`"Reached max session turns for this session"` changes, then this stops working, right?

jerop (changes requested):
would we need such error handling for other fatal errors? is there a way to generalize this to other fatal errors?

action.yml
leehagoodjames: Yes that is correct. I see two approaches:

1. Hitting the `maxSessionTurns` limit causes `gemini-cli` to exit with a specific error code  
  - Pros
    - Very easy to parse downstream  
  - Cons
    - Likely not appropriate for all gemini-cli invocations
2. (Current behavior) Hitting the `maxSessionTurns` limit causes `gemini-cli` to exit with a `0`.    
  - Pros
    - Does not build into `gemini-cli` the assumption that hitting the `maxSessionTurns` indicates that the workload failed
  - Cons
    - Difficult for clients to reliably determine if the `maxSessionTurns` limit was hit, as checking the output errors requires fragile string matching
    
Does gemini-cli currently support any type of output that enumerates the possible reasons for ending ?

sethvargo: I think it would be better for the gemini-cli to publish specific exit codes for these kinds of things, no?

jerop: Agree, this should be an upstream change in Gemini CLI
- https://github.com/google-gemini/gemini-cli/issues/6745
- https://github.com/google-github-actions/run-gemini-cli/issues/217

jerop:
@gemini-cli /review how can we improve the prompt?

jerop:
fyi above workflow failed because the PR is from a fork: https://github.com/google-github-actions/run-gemini-cli/issues/167

jerop:
Closed in favor of using error codes: https://github.com/google-github-actions/run-gemini-cli/issues/217

Issue #153, opened 2025-08-08
Inject ENV vars into prompt


leehagoodjames:
### TL;DR

Numerous issues seem related to the agent not fetching values from environment variables. Rather than only telling the agent to access them, these should be injected into the prompt.

Issue #151, opened 2025-08-08
pr-review: model confusion about line numbers and old/new parts of diffs


dan-ri:
### TL;DR

In the `pr-review` workflow the model frequently:
- Mentions line numbers that are incorrect
- Posts review comments on the wrong lines for what the comment is about
- Includes a `Suggested change` in a review comment that is completely wrong and unrelated to the comment
- Gets confused by diffs and:
  - Reports problems the PR fixes as problems that need to be fixed
  - Says updates need to be made that the PR already makes

### Expected behavior

Review precision with respect to the current state of the branch and all line numbers.

### Action YAML

```yaml
# As in release v0.1.8
```

### Log output

```text

```

leehagoodjames:
Hi @dan-ri, thanks for surfacing this - you are correct on the expected behavior. 

There have been 4 changes to the PR review YAML in the last 24 hours. Can you please include the YAML (or version tag, ex `v0.1.8`) that you are using?


Can you please include any examples of this occurring? It would be helpful to for debugging.

Note: If you aren't using the latest (updated ~1h ago) version of review, can you copy the `/review` workflow from https://github.com/google-github-actions/run-gemini-cli/blob/main/examples/workflows/pr-review/gemini-pr-review.yml and let me know if it resolves any of the issues you are seeing?

dan-ri:
Yes, this is with [v0.1.8.](https://github.com/google-github-actions/run-gemini-cli/releases/tag/v0.1.8)

leehagoodjames:
Thanks for confirming @dan-ri. Can you please provide some info to help debug in the form of:

- Logs including where the agent is planning to post the review
- Screenshots of how this appears on GitHub in the `https://github.com/<OWNER>/<REPO>/pull/<PR_NUMBER>/files` UI?

dan-ri:
I put some work into prompt engineering for this and it's made things significantly better. In the process, though, I've rewritten most of the prompt and configured the action to read it from a separate markdown file.

leehagoodjames:
@dan-ri that is great to hear. Please feel free to suggest that we include these changes into the `main` branch

hackertronix:
@leehagoodjames I am running into the same on v0.1.22

<img width="2168" height="1416" alt="Image" src="https://github.com/user-attachments/assets/7fb2968d-6dfc-4fda-9705-089aed2b8097" />

[...]

Issue #67, opened 2025-07-31
Support separate GCP projects for OpenTelemetry and Inference


jerop:
The script currently sets up one project for both observability and inference: https://github.com/google-github-actions/run-gemini-cli/blob/main/scripts/setup_workload_identity.sh

We should support situations where users have separate projects

PR #212, opened 2025-08-21
feat(actions): use the dispatcher pattern to separate concerns

sethvargo:
This introduces a new top-level workflow "gemini-dispatch" that listens for a series of events. Based on the trigger or contents, it calls another workflow (using `workflow_call`). This helps keep each workflow separate and allows us to re-use workflows in the future.

The generic "invoke" command still needs a lot of work, but this is progress.

Closes https://github.com/google-github-actions/run-gemini-cli/issues/215

/cc @jerop


sethvargo:
The security concern is valid, but this isn't new (and we know it needs fixed already).

.github/workflows/gemini-dispatch.yml
```diff
@@ -0,0 +1,173 @@
...
+            } else {
+              core.setOutput('command', 'fallthrough');
+            }
+
+      - name: 'Acknowledge request'
+        env:
+          GITHUB_TOKEN: '${{ steps.mint_identity_token.outputs.token || secrets.GITHUB_TOKEN || github.token }}'
+          ISSUE_NUMBER: '${{ github.event.pull_request.number || github.event.issue.number }}'
+          MESSAGE: |-
+            🤖 Hi @${{ github.actor }}, I've received your request. I'm working on it now!
```
jerop: it'd be helpful if this message included a link to the workflow run logs

.github/workflows/gemini-invoke.yml
```diff
@@ -107,17 +52,18 @@ jobs:
           git config --global user.name 'gemini-cli[bot]'
           git config --global user.email 'gemini-cli[bot]@users.noreply.github.com'
 
-      - name: 'Run Gemini'
+      - name: 'Run Gemini CLI'
         id: 'run_gemini'
-        uses: './'
+        uses: 'google-github-actions/run-gemini-cli@main' # ratchet:exclude
```
jerop, line 41: this is good for consistency with examples too, but i think it makes it hard to test the changes in `action.yml` within a PR before merging the change - but i suppose one can temporarily use `./` in a branch when testing with workflow dispatch? 🤔

.github/workflows/gemini-invoke.yml
```diff
@@ -130,11 +76,57 @@ jobs:
           gemini_debug: '${{ fromJSON(vars.DEBUG || vars.ACTIONS_STEP_DEBUG || false) }}'
           settings: |-
             {
-              "maxSessionTurns": 50,
+              "maxSessionTurns": 25,
               "telemetry": {
-                "enabled": true,
+                "enabled": ${{ vars.GOOGLE_CLOUD_PROJECT != '' }},
```
jerop, line 66: this is assuming that usage of google cloud includes both inference and observability - https://github.com/google-github-actions/run-gemini-cli/issues/67

.github/workflows/gemini-invoke.yml
```diff
@@ -130,11 +76,57 @@ jobs:
...
+                  "includeTools": [
+                    "add_issue_comment",
+                    "get_issue",
+                    "get_issue_comments",
+                    "list_issues",
+                    "search_issues",
+                    "create_pull_request",
+                    "get_pull_request",
+                    "get_pull_request_comments",
+                    "get_pull_request_diff",
```
jerop, line 89: curious to see if this solves the diff view and line number issues in https://github.com/google-github-actions/run-gemini-cli/issues/151

.github/workflows/gemini-invoke.yml
```diff
@@ -145,20 +137,21 @@ jobs:
...
-            4. Run: echo "${ISSUE_NUMBER}" to get the PR or Issue number
-            5. Run: echo "${REPOSITORY}" to get the github repository in <OWNER>/<REPO> format
-            6. Run: echo "${USER_REQUEST}" to get the user's request
+            1. Run: echo "${TITLE}" to get a title of the pull request or issue
+            2. Run: echo "${DESCRIPTION}" to get a description of the pull request or issue
+            3. Run: echo "${EVENT_NAME}" to learn what kind of GitHub event triggered this request
+            4. Run: echo "${IS_PULL_REQUEST}" to learn whether this is a Pull Request (PR) or Issue
+            5. Run: echo "${ISSUE_NUMBER}" to get the PR or Issue number
+            6. Run: echo "${REPOSITORY}" to get the github repository in <OWNER>/<REPO> format
+            7. Run: echo "${ADDITIONAL_CONTEXT}" to get the user's request and additional context
```
jerop, line 131: a future next step could be to inject the env vars as context instead of asking the model to echo - https://github.com/google-github-actions/run-gemini-cli/issues/153

cc @leehagoodjames

.github/workflows/gemini-dispatch.yml
```diff
@@ -0,0 +1,173 @@
+name: '🔀 Gemini Dispatch'
+
+on:
```
jerop, line 3: should this event be supported for automated issue triage: 

```yaml
 issues:
    types:
      - 'opened'
      - 'reopened'
 ```

**Review by jerop — COMMENTED** [review]
love the simplification of the triggers and context gathering 🎉 

it'd be interesting to extend this to do intelligent routing based on intent using natural language instead of slash commands

File: .github/workflows/gemini-dispatch.yml [hunk 7]
```diff
@@ -0,0 +1,173 @@
...
+          permission-issues: 'write'
+          permission-pull-requests: 'write'
+
+      - name: 'Extract command'
+        id: 'extract_command'
+        uses: 'actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea' # ratchet:actions/github-script@v7
+        env:
+          EVENT_TYPE: '${{ github.event_name }}.${{ github.event.action }}'
+          REQUEST: '${{ github.event.comment.body || github.event.review.body || github.event.issue.body }}'
+        with:
```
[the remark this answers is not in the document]
sethvargo: Disagree, and it was intentional to align the other conditions. There's even a comment about it :)

.github/workflows/gemini-dispatch.yml
sethvargo: Done

.github/workflows/gemini-invoke.yml
sethvargo: If we use `./`, then we _also_ need to add `actions/checkout` first. Then we have two options:

1. Always checkout the main branch, which is equivalent to what you see here
2. Check out the PR branch, but then that introduces an attack path for someone to modify the action and/or prompt that executes

I couldn't think of a great solution here unfortunately. What I've been doing is testing locally in a private repo where there aren't security risks since I'm the only contributor.

.github/workflows/gemini-invoke.yml
sethvargo: Is there a better variable/heuristic to switch on? I found the logs are quite spammy if telemetry is enabled, but the configuration is missing. This was an attempt to fix that by disabling the telemetry if one of the variables is missing.

.github/workflows/gemini-invoke.yml
sethvargo: Maybe, but I think the model is getting confused on LEFT vs RIGHT diff. Or it's just bad at numbers 😄.

.github/workflows/gemini-invoke.yml
sethvargo: I was going to try and rewrite this prompt entirely, but the diff was getting to big. If we get these changes in, then I can take a stab at contextualizing a lot of this.

.github/workflows/gemini-review.yml
```diff
@@ -0,0 +1,267 @@
...
+            - Retrieve the GitHub repository name from the environment variable "${REPOSITORY}".
+            - Retrieve the GitHub pull request number from the environment variable "${PULL_REQUEST_NUMBEER}".
+            - Retrieve the additional user instructions and context from the environment variable "${ADDITIONAL_CONTEXT}".
+            - Use `mcp__github__get_pull_request` to get the title, body, and metadata about the pull request.
+            - Use `mcp__github__get_pull_request_files` to get the list of files that were added, removed, and changed in the pull request.
+            - Use `mcp__github__get_pull_request_diff` to get the diff from the pull request. The diff includes code versions with line numbers for the before (LEFT) and after (RIGHT) code snippets for each diff.
+
+
+            ## Execution Workflow
+
```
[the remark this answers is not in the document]
sethvargo: Done

.github/workflows/gemini-dispatch.yml
sethvargo: Done

.github/workflows/gemini-invoke.yml
jerop: makes sense, will add instructions in our contributing guide about testing in private repos 🙏🏾

.github/workflows/gemini-invoke.yml
jerop: this was just an fyi, as a note for myself to probably provide a separate variable for `OTLP_GOOGLE_CLOUD_PROJECT`

.github/workflows/gemini-invoke.yml
jerop: sounds great, thank you!

jerop (approved):
🚀

.github/workflows/gemini-invoke.yml
jerop: cc @leehagoodjames

PR #232, opened 2025-08-22
chore(ci): refactor scheduled triage workflow

sethvargo:
This refactors the scheduled triage workflow to:

- Only run a single search query (instead of two + munging)
- Operate on 100 issues at a time
- Ensure that the labels that the LLM suggested are actually valid on the repo before applying
- Separates the permissions to modify labels into a dedicated job

There's also a small enhancement of making this workflow run whenever it's changed in a PR, which is nice for devex and serves as a faux integration test.


jerop (approved):

sethvargo:
Oops - there were a few bugs. Let me clean those up: https://github.com/google-github-actions/run-gemini-cli/pull/233

Issue #175, opened 2025-08-13
PR Reviews are not happening automatically


leehagoodjames:
### TL;DR

PR reviews should happen automatically, and they are not. 

### Expected behavior

When a PR is created by a maintainer, owner, or contributor - the PR should be automatically reviewed by gemini.

### Observed behavior

The GitHub action is being skipped

### Action YAML

```yaml
https://github.com/google-github-actions/run-gemini-cli/blob/7c76b68baeb0a33eb3d516cd9fa9242a109054dc/.github/workflows/gemini-pr-review.yml
```

### Log output

```text

```

PR #234, opened 2025-08-25
Auto trigger PR reviews when they are from branches in repo

jerop:
Pull request events identify members as "contributors", so the workflows are not auto-triggered. But there are some contributors who are not members, so workflows would fail on forks.

In this change, we switch the condition for PR events to check for whether the PR was created from a branch in the repo or from a fork.

Note that only users with write access to the repo can create branches.

Fixes https://github.com/google-github-actions/run-gemini-cli/issues/175

cc @leehagoodjames @sethvargo


jerop:
@sethvargo added inline documentation, thanks!

Issue #235, opened 2025-08-25
Synchronize examples/workflows with .github/workflows


jerop:
We have made improvements to our own workflows that we'd like to share with other users

PR #238, opened 2025-08-25
rename workflow from `gemini-issue-scheduled-triage` to `gemini-scheduled-triage`

jerop:
- this makes workflow consistent with the other workflows
- this makes workflow names simpler


sethvargo (approved):
LGTM, but I'd like to actually move this into gemini-dispatch somehow (I haven't figured out how yet)

jerop:
sounds good, working updating example workflows so we can cut a release with current state - https://github.com/google-github-actions/run-gemini-cli/issues/235

PR #244, opened 2025-08-25
chore: script to generate examples

### LINKED ISSUE #68: Make Workflow Config DRY
### TL;DR

Workflows are identically defined in `.github/workflows/` and in `workflows/<workflow_identifer>/` folders.

This issue tracks the work to automatically link the two files, through either automation, symlinks, or something else. 

Requirements:
- (P0) Local development and testing of files in `.github/workflows/` should not have any regressions, even when using  [nektos/act](https://github.com/nektos/act) for testing.
- (P1) The diffs in code reviews should only be for one file, since they are clones

### Detailed design

_No response_

### Additional information

_No response_

leehagoodjames:
This PR fixes #68.

It introduces a script to automatically generate the example workflow files from the main workflow files.
This will help to keep the workflows DRY and avoid duplication.

### DIFF
--- a/.github/workflows/gemini-triage.yml
+++ b/.github/workflows/gemini-triage.yml
@@ -69,6 +69,7 @@ jobs:
           gcp_service_account: '${{ vars.SERVICE_ACCOUNT_EMAIL }}'
           gemini_api_key: '${{ secrets.GEMINI_API_KEY }}'
           use_vertex_ai: '${{ vars.GOOGLE_GENAI_USE_VERTEXAI }}'
+          google_api_key: '${{ secrets.GOOGLE_API_KEY }}'
           use_gemini_code_assist: '${{ vars.GOOGLE_GENAI_USE_GCA }}'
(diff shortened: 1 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -65,6 +65,7 @@ Before contributing, ensure you have:
       ```sh
       npm run docs
       ```
+    - If you update workflow files in `/.gemini/workflows/`, run `./scripts/generate-examples.sh` to auto-generate the examples.
```
sethvargo, line 68: I wonder if we should do this as part of the release workflow too?

sethvargo (approved):

CONTRIBUTING.md
leehagoodjames: That would be great. Automation >> readme steps

leehagoodjames: We can also add a disclaimer header `This is an autogenerated file ...`, but I didn't do that since we encourage people to copy from the examples

Issue #257, opened 2025-09-01
Workload identity provider authentication ignored


kuisathaverat:
### TL;DR

I have a Workload Identity Provider (WIP) configuration that works with `google-github-actions/auth@v2`, I have tried to use the same WIP with `google-github-actions/run-gemini-cli@main`, and it seems to be ignored. It tries to use the API token.

### Expected behavior

I added the following roles to the service account we use for our WIP (we use it with `google-github-actions/auth@v2` daily)

Configuring the WIP we use with `google-github-actions/auth@v2` daily, and updating the service account we use with the following roles:

```
roles/logging.logWriter: To write logs to Cloud Logging.
roles/monitoring.editor: To write metrics to Cloud Monitoring.
roles/cloudtrace.agent: To send traces to Cloud Trace.
roles/aiplatform.user: To make inference calls to Vertex AI.
roles/cloudaicompanion.user: To make inference calls using Gemini Code Assist.
roles/iam.serviceAccountTokenCreator: To generate access tokens.
```

And running the action with the WIP configuration. The action should run the prompt without errors.

### Observed behavior

The Gemini CLI tries to use the API Key instead of the Google Credentials.
It creates the credentials files, but for some reason, the action does not use them.

### Action YAML

```yaml
- name: Gemini CLI
        uses: google-github-actions/run-gemini-cli@main
        id: gemini-cli-run
        with:
          gemini_debug: "true"
          gcp_workload_identity_provider: '${{ secrets.WORKLOAD_IDENTITY_PROVIDER_ID }}'
          gcp_service_account: '${{ secrets.SERVICE_ACCOUNT_EMAIL }}'
          gcp_project_id: '${{ vars.GCP_PROJECT }}'
          prompt: "hi"
```

### Log output

```text
2025-09-01T14:35:40.0430394Z ##[group]Run google-github-actions/run-gemini-cli@v0
2025-09-01T14:35:40.0430799Z with:
2025-09-01T14:35:40.0431398Z   gcp_workload_identity_provider: ***
2025-09-01T14:35:40.0431882Z   gcp_service_account: ***
2025-09-01T14:35:40.0432113Z   gcp_project_id: my-project
2025-09-01T14:35:40.0432649Z   prompt: 
hi

2025-09-01T14:35:40.0433186Z   use_vertex_ai: false
2025-09-01T14:35:40.0433396Z   use_gemini_code_assist: false
[...]

Issue #258, opened 2025-09-03
run-gemini-cli fails with Cloud Code Private API error when using Code Assist and WIF


grhoades:
### **Overview**

When attempting to use the `run-gemini-cli` action with `use_gemini_code_assist: true` and authenticating via Workload Identity Federation, the action fails during the `gemini-cli` execution. The error message indicates that the "Cloud Code Private API" is not enabled, even though all required public APIs and IAM permissions appear to be correctly configured.

### **Workflow Configuration**

Here is the relevant step from our workflow file. Project-specific identifiers have been replaced with placeholders.

```yaml
- name: 'Run Gemini issue analysis'
  id: 'gemini_analysis'
  uses: 'google-github-actions/run-gemini-cli@v0'
  with:
    gcp_workload_identity_provider: 'projects/[PROJECT_NUMBER]/locations/global/workloadIdentityPools/[POOL_ID]/providers/[PROVIDER_ID]'
    gcp_project_id: '[YOUR_GCP_PROJECT_ID]'
    gcp_location: 'us-central1'
    gcp_service_account: '[YOUR_SERVICE_ACCOUNT_EMAIL]'
    use_gemini_code_assist: true
    settings: |
      {
        "maxSessionTurns": 25,
        "telemetry": { "enabled": true, "target": "gcp" },
        "coreTools": [ "run_shell_command(echo)" ]
      }
    prompt: |
      ## Role
      You are an issue triage assistant...
```

### Expected behavior

Workflows run correctly and utilize the Google Cloud Code Assist Enterprise subscription.

### **Error Log**

The workflow fails with the following error. The project number in the URL has been redacted.

```
##[error]An unexpected critical error occurred:
An unexpected critical error occurred:
Error: Cloud Code Private API has not been used in project [YOUR_PROJECT_NUMBER] before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/cloudcode-pa.googleapis.com/overview?project=[YOUR_PROJECT_NUMBER] then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.
    at Gaxios._request (/usr/local/lib/node_modules/@google/gemini-cli/node_modules/gaxios/build/src/gaxios.js:142:23)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    # ... stack trace continues
```

### **Troubleshooting Steps Performed**

We have already attempted the following steps without success:
[...]

PR #289, opened 2025-09-17
feat: add auth input validation

jerop:
Adds a validation step to the action to ensure that authentication inputs are configured correctly. This prevents common misconfigurations and improves security.

The validation is performed by a new script, `scripts/validate-inputs.sh`, which is called as the first step in the action.

Key changes:
- A new `validate-inputs.sh` script is added to check authentication configuration.
- The `action.yml` is updated to call this script.
- Follows principle of least privilege by using boolean flags to indicate whether inputs are set.
- The validation enforces that exactly one authentication method is used, providing clear error messages to the user if the configuration is invalid. This makes the action more robust, secure, and easier to debug.

Fixes https://github.com/google-github-actions/run-gemini-cli/issues/257 https://github.com/google-github-actions/run-gemini-cli/issues/258

### DIFF
--- a/scripts/validate-inputs.sh
+++ b/scripts/validate-inputs.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+set -euo pipefail
+
+# Auth inputs (as boolean presence flags)
+gemini_api_key_present="${INPUT_GEMINI_API_KEY_PRESENT:-false}"
+google_api_key_present="${INPUT_GOOGLE_API_KEY_PRESENT:-false}"
+gcp_workload_identity_provider_present="${INPUT_GCP_WORKLOAD_IDENTITY_PROVIDER_PRESENT:-false}"
+gcp_project_id_present="${INPUT_GCP_PROJECT_ID_PRESENT:-false}"
+gcp_service_account_present="${INPUT_GCP_SERVICE_ACCOUNT_PRESENT:-false}"
+
+# Other inputs (values needed)
+use_vertex_ai="${INPUT_USE_VERTEX_AI:-false}"
+use_gemini_code_assist="${INPUT_USE_GEMINI_CODE_ASSIST:-false}"
+
+# Count number of auth methods
+auth_methods=0
+if [[ "${gemini_api_key_present}" == "true" ]]; then ((auth_methods++)); fi
+if [[ "${google_api_key_present}" == "true" ]]; then ((auth_methods++)); fi
+if [[ "${gcp_workload_identity_provider_present}" == "true" ]]; then ((auth_methods++)); fi
+
+if [[ ${auth_methods} -eq 0 ]]; then
+  echo "::error title=Configuration error::No authentication method provided. Please provide one of 'gemini_api_key', 'google_api_key', or 'gcp_workload_identity_provider'." >&2
+  exit 1
+fi
+
+if [[ ${auth_methods} -gt 1 ]]; then
+  echo "::error title=Configuration error::Multiple authentication methods provided. Please use only one of 'gemini_api_key', 'google_api_key', or 'gcp_workload_identity_provider'." >&2
+  exit 1
+fi
+
+# WIF validation
+if [[ "${gcp_workload_identity_provider_present}" == "true" ]]; then
+  if [[ "${gcp_project_id_present}" != "true" || "${gcp_service_account_present}" != "true" ]]; then
+    echo "::error title=Configuration error::When using Workload Identity Federation ('gcp_workload_identity_provider'), you must also provide 'gcp_project_id' and 'gcp_service_account'." >&2
+    exit 1
+  fi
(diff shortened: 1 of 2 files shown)


jerop:
@gemini-cli /review

scripts/validate-inputs.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+set -euo pipefail
+
+# Auth inputs (as boolean presence flags)
+gemini_api_key_present="${INPUT_GEMINI_API_KEY_PRESENT:-false}"
+google_api_key_present="${INPUT_GOOGLE_API_KEY_PRESENT:-false}"
+gcp_workload_identity_provider_present="${INPUT_GCP_WORKLOAD_IDENTITY_PROVIDER_PRESENT:-false}"
+gcp_project_id_present="${INPUT_GCP_PROJECT_ID_PRESENT:-false}"
+gcp_service_account_present="${INPUT_GCP_SERVICE_ACCOUNT_PRESENT:-false}"
```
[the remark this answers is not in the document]
jerop: there are no backslashes?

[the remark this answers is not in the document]
jerop: double quotes is required by out linting -- will add this to GEMINI.md

scripts/validate-inputs.sh
```diff
@@ -0,0 +1,65 @@
...
+
+# Other inputs (values needed)
```
leehagoodjames, line 11: How are these inputs different from the "Auth inputs" above?

scripts/validate-inputs.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+set -euo pipefail
+
+# Auth inputs (as boolean presence flags)
+gemini_api_key_present="${INPUT_GEMINI_API_KEY_PRESENT:-false}"
+google_api_key_present="${INPUT_GOOGLE_API_KEY_PRESENT:-false}"
+gcp_workload_identity_provider_present="${INPUT_GCP_WORKLOAD_IDENTITY_PROVIDER_PRESENT:-false}"
```
leehagoodjames, line 7: Are these all auth keys?

scripts/validate-inputs.sh
```diff
@@ -0,0 +1,65 @@
...
+fi
+
+# WIF validation
+if [[ "${gcp_workload_identity_provider_present}" == "true" ]]; then
+  if [[ "${gcp_project_id_present}" != "true" || "${gcp_service_account_present}" != "true" ]]; then
+    echo "::error title=Configuration error::When using Workload Identity Federation ('gcp_workload_identity_provider'), you must also provide 'gcp_project_id' and 'gcp_service_account'." >&2
+    exit 1
+  fi
+  if [[ "${use_vertex_ai}" != "true" && "${use_gemini_code_assist}" != "true" ]]; then
+    echo "::error title=Configuration error::When using Workload Identity Federation, you must set either 'use_vertex_ai' or 'use_gemini_code_assist' to 'true'." >&2
```
leehagoodjames, line 38: Nit: make the current state clear
```suggestion
    echo "::error title=Configuration error:: 'use_vertex_ai' and 'use_gemini_code_assist' are both set to 'false'. When using Workload Identity Federation, you must set either 'use_vertex_ai' or 'use_gemini_code_assist' to 'true'." >&2
```

scripts/validate-inputs.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+set -euo pipefail
+
```
leehagoodjames, line 3: Please add a comment that says if validation failure exclusively is for failing fast, because failures in validation will definitely **fail downstream**, or if some of these checks are to ensure **best practices** that will not necessarily cause downstream failures.

galz10 (approved):

leehagoodjames (commented):
This looks good. 

It is unclear if all of these are examples of failing fast, or if some of these checks will cause existing user's workflows - which are WAI but configured in a way that would fail this validation - to suddenly start failing.

scripts/validate-inputs.sh
jerop: They're not auth keys, they are booleans indicating whether auth keys are available or not

scripts/validate-inputs.sh
jerop: yes, they are different because it's the actual input as opposed to a flag indicating whether an input is available

scripts/validate-inputs.sh
jerop: starting with warnings: https://github.com/google-github-actions/run-gemini-cli/pull/300

scripts/validate-inputs.sh
jerop: addressed in https://github.com/google-github-actions/run-gemini-cli/pull/297

PR #302, opened 2025-09-18
feat: add command substitution security warning

jerop:
Adds a security warning to all Gemini CLI workflow prompts, instructing the model to avoid using command substitution features like `$(...)`, `<(...)`, or `>(...)` in shell commands.

This is a preventative measure to avoid errors where the Gemini CLI rejects commands that use command substitution for security reasons. This change makes the security constraints of the tool explicit to the model.


jerop:
@gemini-cli /review

jerop:
@gemini-cli /review

jerop:
@gemini-cli /review

PR #329, opened 2025-09-24
feat: Optionally install extensions

CallumHYoung:
**Summary**

This PR introduces the optional installation of extensions

**Changes**
- New `extensions` input: A newline separated list of extension repository URLs.
- Conditional installation: The install workflow checks the extension flag and installs extensions that are listed

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,9 +188,9 @@ go to the [Gemini Assistant workflow documentation](./examples/workflows/gemini-
 
 <!-- BEGIN_AUTOGEN_OUTPUTS -->
 
--   <a name="__output_summary"></a><a href="#user-content-__output_summary"><code>summary</code></a>: The summarized output from the Gemini CLI execution.
+-   `summary`: The summarized output from the Gemini CLI execution.
 
--   <a name="__output_error"></a><a href="#user-content-__output_error"><code>error</code></a>: The error output from the Gemini CLI execution, if any.
+-   `error`: The error output from the Gemini CLI execution, if any.
 
 
 <!-- END_AUTOGEN_OUTPUTS -->


CallumHYoung:
I've updated the PR to:

* Instead use an input 'extensions' which accepts [multiple new-line separated extensions urls](https://github.com/CallumHYoung/testrepo/blob/2181946a457f22dddedd7b687acecde934f4df85/.github/workflows/gemini-review.yml#L65)
* Iterate over `extensions` and [runs each individually](https://github.com/CallumHYoung/run-gemini-cli/blob/750c8efb8a85309d16ff108230fd68f122c749cf/action.yml#L231), additionally confirming to the consent check to install the extension

Here is confirmation of it working in the logs [here](https://github.com/CallumHYoung/testrepo/actions/runs/18016146486/job/51261452591), installing multiple listed extensions (in the format I linked in the first bullet point)

action.yml
```diff
@@ -68,6 +68,9 @@ inputs:
       For more information, see the [Gemini CLI documentation](https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/authentication.md).
     required: false
     default: 'false'
+  extensions:
+    description: 'A newline-separated list of Gemini CLI extensions to install.'
+    required: false
```
jerop, line 73: could we make this an array instead?

action.yml
```diff
@@ -220,6 +225,17 @@ runs:
           echo "Error: Gemini CLI not found in PATH"
           exit 1
         fi
+        if [[ -n '${{ inputs.extensions }}' ]]; then
```
verbanicm: this should use the environment variable instead of interpolation

CallumHYoung: Done!

action.yml
CallumHYoung: Sure, done!

CallumHYoung:
Here is the updated test with the suggested changes (Extensions as a JSON array, using EXTENSIONS env variable)

https://github.com/CallumHYoung/testrepo/actions/runs/18020529034/job/51276560027

[Extensions in test repository](https://github.com/CallumHYoung/testrepo/blob/97b0a2d9eb2171eadc47080ac49e5a7175024309/.github/workflows/gemini-review.yml#L65)

action.yml
jerop: Is it actually possible to make it an input of type array? Have you tested this out?

https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax#inputs

action.yml
jerop: can this line (227) also use the environment variable?

CallumHYoung: I think this is addressed this with 41ea373, let me know if I'm not understanding correctly

action.yml
CallumHYoung: Can you clarify what you mean by 'type array'? My understanding is that everything is passed as a string, and I've tested this as an array in JSON

https://github.com/google-github-actions/run-gemini-cli/pull/329#issuecomment-3335285601 shows some testing, and I've also attempted to call /security:analyze (the extensions command) in two different workflows

Here's the run with just using /security:analyze as the prompt https://github.com/CallumHYoung/testrepo/actions/runs/18104703538

Heres the run with the comments from asking gemini to make the extension call and then using the analysis to comment to the PR. https://github.com/CallumHYoung/testrepo/actions/runs/18105319495

README.md
```diff
@@ -179,16 +179,18 @@ go to the [Gemini Assistant workflow documentation](./examples/workflows/gemini-
...
 
 ### Outputs
 
 <!-- BEGIN_AUTOGEN_OUTPUTS -->
 
--   `summary`: The summarized output from the Gemini CLI execution.
+-   <a name="__output_summary"></a><a href="#user-content-__output_summary"><code>summary</code></a>: The summarized output from the Gemini CLI execution.
 
--   `error`: The error output from the Gemini CLI execution, if any.
+-   <a name="__output_error"></a><a href="#user-content-__output_error"><code>error</code></a>: The error output from the Gemini CLI execution, if any.
```
jerop: please undo these changes

CallumHYoung: Done!

jerop (approved):

PR #346, opened 2025-10-06
feat(issue triage): Get all labels of the repository

chouetz:
Use a loop on the paginated API to collect all the labels of the repo, as the current implementation only get 30 of them. 

I did not modified the current workflows you are using [here](https://github.com/google-github-actions/run-gemini-cli/blob/main/.github/workflows/gemini-triage.yml), but if it makes sense for you I can update the PR.

We might think about an additional option to use specific patterns to select only some labels lately?

Thanks in advance for the review


chouetz:
> @chouetz thanks for the contribution - please make the same change in `.github/workflows` so we can dogfood this 🙏🏾

@jerop Done, thanks for the review!

PR #348, opened 2025-10-09
Move `gemini-invoke` to custom command.

joshualitt:
For simplicity, this PR only moves `gemini-invoke` to a custom command. The other prompts will be moved in a follow-up.

[Demo](https://github.com/joshualitt/test-repo/issues/1) of this working e2e


.github/commands/gemini-invoke.toml
jerop, line 1: what do you think of putting these commands within `.github/commands` so that the commands are alongside the workflows in `.github` folder

joshualitt: Done!

PR #357, opened 2025-10-14
Fix interpolation syntax.

joshualitt:
A simple PR to correctly interpolate environment variables in a few places.

### DIFF
--- a/.github/commands/gemini-scheduled-triage.toml
+++ b/.github/commands/gemini-scheduled-triage.toml
@@ -6,7 +6,7 @@ You are a highly efficient Issue Triage Engineer. Your function is to analyze Gi
 
 ## Primary Directive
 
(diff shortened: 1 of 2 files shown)


.github/commands/gemini-scheduled-triage.toml
```diff
@@ -6,15 +6,15 @@ You are a highly efficient Issue Triage Engineer. Your function is to analyze Gi
 
 ## Primary Directive
 
-You will retrieve issue data and available labels from environment variables, analyze the issues, and assign the most relevant labels. You will then generate a single JSON array containing your triage decisions and write it to the file path specified by the `${GITHUB_ENV}` environment variable.
+You will retrieve issue data and available labels from environment variables, analyze the issues, and assign the most relevant labels. You will then generate a single JSON array containing your triage decisions and write it to `!{echo $GITHUB_ENV}`.
```
joshualitt, line 9: This one is a judgement call, but I think putting  the file path here will be an improvement, vs. just putting `${GITHUB_ENV}` as a literal string.

PR #354, opened 2025-10-11
fix: Adapt to GitHub MCP Tooling Consolidation

cperez08:
This pull request addresses recent workflow failures where the **Gemini CLI could not perform review tasks** due to **breaking changes** in the GitHub MCP server.

These changes were introduced in the GitHub MCP server's latest release, which consolidated the `get_pull_request*` tools into a single `pull_request_read` with multiple methods.

### Key Changes and Fixes

1.  **Dependency Pinning:** The Docker image for the GitHub MCP Server is now **pinned to a specific version (v0.18.0)** instead of using `latest`. This prevents unexpected disruptions from future upstream updates and ensures consistent action stability.
2.  **Tool Reference Update:** Updates MCP tool references and associated prompts for the `review` and `invoke` workflows and examples.

For full details on the upstream changes, see the GitHub MCP Server release notes:
[v0.18.0 Release Notes](https://github.com/github/github-mcp-server/releases/tag/v0.18.0)

Fixes #353

PR #359, opened 2025-10-15
Fix incorrect tool name in gemini-invoke and add a few missing tools.

joshualitt:
"get_pull_request_comments" does not appear to be a valid tool name anymore. I think this is now `get_pull_request_review_comments`. There are also a few pull request tools which appear to be missing, so those have been added as well.


joshualitt:
Let's land #354 instead, as it pins to a consistent version.

Issue #360, opened 2025-10-16
Upload telemetry from running actions to Github as artifacts


joshualitt:
Today, it is very difficult to debug these actions. The telemetry logs are uploaded to GCP, but because of the use of the streaming collector, they may be truncated. We should collect telemetry locally, and upload it to GCP ourselves, to ensure we have solid telemetry data, at least on the action itself if not in GCP

PR #361, opened 2025-10-16
Switch to local telemetry and upload manually to GCP

joshualitt:
Resolves https://github.com/google-github-actions/run-gemini-cli/issues/360

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -301,10 +301,10 @@ runs:
     - name: 'Upload Gemini CLI outputs'
       if: |-
         ${{ inputs.upload_artifacts }}
-      uses: 'actions/upload-artifact@v4'
+      uses: 'actions/upload-artifact@v4' # ratchet:exclude
       with:
-        name: gemini-output
-        path: gemini-artifacts/
+        name: 'gemini-output'
+        path: 'gemini-artifacts/'
 
     - name: 'Upload Telemetry to Google Cloud'
       if: |-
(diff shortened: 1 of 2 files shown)


.github/workflows/gemini-invoke.yml
```diff
@@ -67,8 +67,10 @@ jobs:
                 "maxSessionTurns": 25
               },
               "telemetry": {
-                "enabled": ${{ vars.GOOGLE_CLOUD_PROJECT != '' }},
-                "target": "gcp"
+                "enabled": true,
+                "target": "local",
+                "otlpEndpoint": "",
```
jerop: is this necessary?

scripts/collector-gcp.yaml.template
```diff
@@ -1,34 +1,33 @@
...
+
 exporters:
   googlecloud:
-    project: 'OTLP_GOOGLE_CLOUD_PROJECT'
-    metric:
-      prefix: 'custom.googleapis.com/gemini_cli'
+    project: "OTLP_GOOGLE_CLOUD_PROJECT"
     log:
-      default_log_name: 'gemini_cli'
+      default_log_name: "gemini-cli-telemetry"
```
jerop: let's not change the log name because this is also what's used upstream

action.yml
```diff
@@ -307,6 +294,80 @@ runs:
         PROMPT: '${{ inputs.prompt }}'
         GEMINI_MODEL: '${{ inputs.gemini_model }}'
 
+    - name: 'Upload Gemini CLI outputs'
+      uses: 'actions/upload-artifact@v4'
+      with:
+        name: gemini-output
+        path: gemini-artifacts/
```
jerop: are there concerns about all users with read access having access to full logs? https://docs.github.com/en/actions/concepts/workflows-and-actions/workflow-artifacts

cc @cornmander

joshualitt: To unblock this, I've gone ahead and moved the upload behind a flag that defaults to false. We can investigate flipping the default once we've determined if this is safe or not.

joshualitt:
Documentation updated, ptal!

Also confirmed the new environment variable works as expected, at least in triage.

PR #368, opened 2025-10-23
fix(action): correct upload artifacts condition

jerop:
The previous condition `${{ inputs.upload_artifacts }}` would evaluate to true for any non-empty string, causing artifacts to be uploaded even when the input was false.

This change corrects the condition to `${{ inputs.upload_artifacts == true }}`, ensuring that artifacts are only uploaded when the `upload_artifacts` input is explicitly set to true.

Co-authored-by: calyoung@google.com

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -72,7 +72,7 @@ inputs:
     description: 'A list of Gemini CLI extensions to install.'
     required: false
   upload_artifacts:
-    description: 'Whether or not to upload artifacts to the github action.'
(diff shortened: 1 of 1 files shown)


jerop:
this is a follow up to https://github.com/google-github-actions/run-gemini-cli/pull/361 - cc @joshualitt

jerop:
this supersedes https://github.com/google-github-actions/run-gemini-cli/pull/367 by @CallumHYoung

PR #384, opened 2025-11-03
Get information about the workflows triggerring Gemini CLI

MJjainam:

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ go to the [Gemini Assistant workflow documentation](./examples/workflows/gemini-
 
 -   <a name="__input_extensions"></a><a href="#user-content-__input_extensions"><code>extensions</code></a>: _(Optional)_ A list of Gemini CLI extensions to install.
 
--   <a name="__input_upload_artifacts"></a><a href="#user-content-__input_upload_artifacts"><code>upload_artifacts</code></a>: _(Optional, default: `false`)_ Whether or not to upload artifacts to the github action.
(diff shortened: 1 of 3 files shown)


jerop (changes requested):
could we be consistent with the name; like do we retain the `gemini-` prefix or not? also do we want to keep the name of the file

examples/workflows/gemini-assistant/gemini-invoke.yml
```diff
@@ -49,6 +49,7 @@ jobs:
           ISSUE_NUMBER: '${{ github.event.pull_request.number || github.event.issue.number }}'
           REPOSITORY: '${{ github.repository }}'
           ADDITIONAL_CONTEXT: '${{ inputs.additional_context }}'
+          GHA_NAME: 'invoke_gemini_cli'
```
jerop: I wonder if `GH_WORKFLOW_NAME` would be better, these are workflows not actions

MJjainam: I am using the workflow name itself - this would reduce handling hardcoded names.

examples/workflows/gemini-assistant/gemini-invoke.yml
```diff
@@ -25,6 +25,15 @@ jobs:
       issues: 'write'
       pull-requests: 'write'
     steps:
+      - name: 'Sanitize workflow name'
+        id: 'sanitize_workflow_name'
+        env:
+          RAW_WORKFLOW_NAME: ${{ github.workflow }}
```
jerop:
```suggestion
          RAW_WORKFLOW_NAME: '${{ github.workflow }}'
```

PR #390, opened 2025-11-12
feat(telemetry): set GHA_NAME dynamically from workflow

MJjainam:
Sets the GHA_NAME environment variable for the Gemini CLI action to the name of the executing workflow. This makes the telemetry data more accurate and easier to filter. A new step is added to sanitize the workflow name by removing emojis and leading/trailing whitespace, ensuring the GHA_NAME is a clean string.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -87,8 +87,8 @@ inputs:
     description: 'Whether or not to use pnpm instead of npm to install gemini-cli'
     required: false
     default: 'false'
-  raw_workflow_name:
-    description: 'The raw GitHub workflow name, used for telemetry purposes.'
+  workflow_name:
+    description: 'The GitHub workflow name, used for telemetry purposes.'
     required: false
     default: '${{ github.workflow }}'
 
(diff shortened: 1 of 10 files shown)


action.yml
```diff
@@ -156,6 +156,16 @@ runs:
         INPUT_USE_VERTEX_AI: '${{ inputs.use_vertex_ai }}'
         INPUT_USE_GEMINI_CODE_ASSIST: '${{ inputs.use_gemini_code_assist }}'
 
+    - name: 'Sanitize workflow name'
+      id: 'sanitize_workflow_name'
+      shell: 'bash'
```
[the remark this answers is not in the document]
MJjainam: @gemini-cli why is there a space in the sed command?

examples/workflows/gemini-assistant/gemini-invoke.yml
```diff
@@ -62,6 +62,7 @@ jobs:
           use_gemini_code_assist: '${{ vars.GOOGLE_GENAI_USE_GCA }}'
           use_vertex_ai: '${{ vars.GOOGLE_GENAI_USE_VERTEXAI }}'
           upload_artifacts: '${{ vars.UPLOAD_ARTIFACTS }}'
+          raw_workflow_name: 'gemini-invoke'
```
jerop: shouldn't this be an env var instead of an input?

jerop: same for the others below

jerop: resolved offline

MJjainam: github.workflow would print the root workflow which is dispatch in our case. We will have to explicitly pass the workflow name and use github.workflow only for default use cases.

MJjainam:
@gemini-cli  /review

MJjainam:
@gemini-cli /review

jerop (approved):

Issue #217, opened 2025-08-21
Leverage Gemini CLI Exit Codes for Robust Error Handling


jerop:
## Problem

Currently, GitHub Action has to rely on fragile string matching to detect specific error conditions from gemini-cli output (see #205, #206). This approach is brittle and breaks when error messages change.

## Solution

Use exit codes once they're available in Gemini CLI - https://github.com/google-gemini/gemini-cli/issues/6745.

Example usage of error codes:

| Exit Code | Error Type | Action Annotation | Retry? | Cleanup Action |
|-----------|------------|------------------|---------|----------------|
| 0 | Success | None | No | Continue workflow |
| 2 | Auth Failure | `::error` | No | Post auth help comment |
| 3 | Rate Limit | `::warning` | Yes (3x) | Wait with backoff |
| 4 | Session Limit | `::error` | No | Post session limit comment |
| 5 | Network Issues | `::warning` | Yes (2x) | Retry with delay |
| 6 | Config Error | `::error` | No | Post config help comment |
| 7 | Permission Denied | `::error` | No | Post permission help comment |
| Other | General Failure | `::error` | No | Post generic error comment |

jakemac53:
@jerop looks like this is unblocked, are you actively working on it or would you like me to pick it up?

jakemac53:
Actually nvm it looks like maybe https://github.com/google-github-actions/run-gemini-cli/pull/206 should just be directly updated

jerop:
@jakemac53 thanks for volunteering to take this on - @leehagoodjames have you started working on updating #206 to use the error codes, or is this something @jakemac53 can help with?

megha1188:
I updated gemini cli to return specific error codes here https://github.com/google-gemini/gemini-cli/pull/13728

megha1188:
@jerop  and @leehagoodjames 
Hi,

I updated the error codes that gemini cli return with this PR https://github.com/google-gemini/gemini-cli/pull/13728.

Now looking at how github actions should respond to these error codes? How should we handle these error codes in github actions from the product point of view?

MJjainam:
The Exit Codes of Gemini CLI is taken care as part of the PR: https://github.com/google-gemini/gemini-cli/pull/13728

PR #401, opened 2025-11-26
Updated exit codes with clearer message.

megha1188:
This PR improves error handling by providing more specific error messages based on the exit code from the Gemini CLI.

Previously, the action would exit with a generic "Gemini CLI execution failed" message. This change introduces a case statement that maps common exit codes to user-friendly messages, making it easier to diagnose the root cause of a failure. The action now also exits with the original exit code from the Gemini CLI, which can be useful for scripting and automation.

For example, a failure due to an invalid API key will now produce the message: "Gemini CLI execution failed (Authentication failure. Please check your credentials (API key or Google Cloud auth))" and exit with code 41.

Related Issue #217

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -83,6 +83,10 @@ inputs:
     description: 'Whether to upload artifacts to the github action.'
     required: false
     default: 'false'
+  use_pnpm:
+    description: 'Whether or not to use pnpm instead of npm to install gemini-cli'
+    required: false
+    default: 'false'
 
 outputs:
   summary:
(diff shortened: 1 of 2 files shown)


action.yml
```diff
@@ -323,10 +326,40 @@ runs:
...
+              MSG="User cancelled. The operation was interrupted."
+              ;;
+            *)
+              MSG="Unknown error (exit code: ${EXIT_CODE})"
+              ;;
+          esac
+          
           LAST_LINE="$(tail -n1 "${TEMP_STDERR}")"
-          echo "::error title=Gemini CLI execution failed::${LAST_LINE}"
+          echo "::error title=Gemini CLI execution failed (${MSG})::${LAST_LINE}"
```
MJjainam: This line would not be concatenated in correct format.

MJjainam:
@gemini-cli /review

MJjainam (approved):

PR #402, opened 2025-11-28
fix #371 use JSON output of gemini cli

### LINKED ISSUE #371: Use structured JSON output
JSON output in headless mode returns structured data including response, statistics, and metadata. This format is ideal for programmatic processing and automation scripts.

We can use this to show the stats after a job is completed, otherwise show error message.

gsehgal:
Fixes #371 
Using JSON output of Gemini CLI 

Test Scenarios:

1. Created an ISSUE which caused failure and verified the output to be in json format:

image

2. Created a PULL request and verified the positive response was on JSON format.

image

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -337,7 +337,7 @@ runs:
           echo "${RESPONSE}" >> "${GITHUB_OUTPUT}"
         else
           cat "${TEMP_STDOUT}" >> "${GITHUB_OUTPUT}"
-        fi  
+        fi
         echo "EOF" >> "${GITHUB_OUTPUT}"
 
         # Set the captured errors as a step output, supporting multiline


MJjainam:
@gemini-cli /review

action.yml
```diff
@@ -289,16 +289,14 @@ runs:
         # Keep track of whether we've failed
         FAILED=false
 
-        # Run Gemini CLI with the provided prompt, streaming responses in debug
+        # Run Gemini CLI with the provided prompt, using JSON output format
         if [[ "${DEBUG}" = true ]]; then
           echo "::warning::Gemini CLI debug logging is enabled. This will stream responses, which could reveal sensitive information if processed with untrusted inputs."
-          if ! { gemini --yolo --prompt "${PROMPT}" 2> >(tee "${TEMP_STDERR}" >&2) | tee "${TEMP_STDOUT}"; }; then
```
MJjainam, line 295: Why did we remove this line?

gsehgal: @MJjainam there is no need to call gemini twice, If debug I am just print  warning message.

action.yml
```diff
@@ -312,19 +310,44 @@ runs:
...
+        # Parse JSON output to extract response and errors
+        # If output is not valid JSON, RESPONSE will be empty and we'll rely on stderr for errors
+        if jq -e . "${TEMP_STDOUT}" >/dev/null 2>&1; then
+           RESPONSE=$(jq -r '.response // ""' "${TEMP_STDOUT}")
+           ERROR_JSON=$(jq -c '.error // empty' "${TEMP_STDOUT}")
+        else
+           RESPONSE=""
+           ERROR_JSON=""
+           # If FAILED is false but JSON is invalid, something is wrong
+           if [[ "${FAILED}" = false ]]; then
```
MJjainam: Why are we checking ${FAILED} again? Shouldn't the else run be enough to decide that TEMP_STDOUT is invalid?

action.yml
gsehgal: @MJjainam Have added a DEBUG block to print the contents of  TEMP_$$$$

action.yml
```diff
@@ -312,19 +310,44 @@ runs:
           touch gemini-artifacts/telemetry.log
         fi
 
+        # Parse JSON output to extract response and errors
+        # If output is not valid JSON, RESPONSE will be empty and we'll rely on stderr for errors
+        if jq -e . "${TEMP_STDOUT}" >/dev/null 2>&1; then
+           RESPONSE=$(jq -r '.response // ""' "${TEMP_STDOUT}")
+           ERROR_JSON=$(jq -c '.error // empty' "${TEMP_STDOUT}")
```
MJjainam: Where are we taking care of TEMP_STDERR's ERROR_JSON ?

action.yml
gsehgal: @MJjainam no TEMP_STDOUT Invalid does not mean Gemini CLI failed, it could be a successful run with invalid JSON output.

action.yml
gsehgal: @MJjainam TEMP_STDERR captures errors which will not be part of JSON, things that gemini cli directly writes to stderr and not the final json output.

action.yml
```diff
@@ -312,19 +315,44 @@ runs:
...
+        else
+          cat "${TEMP_STDERR}" >> "${GITHUB_OUTPUT}"
+        fi
         echo "EOF" >> "${GITHUB_OUTPUT}"
 
         if [[ "${FAILED}" = true ]]; then
-          LAST_LINE="$(tail -n1 "${TEMP_STDERR}")"
-          echo "::error title=Gemini CLI execution failed::${LAST_LINE}"
+          # If we have a structured error from JSON, use it for the error message
+          if [[ -n "${ERROR_JSON}" ]]; then
```
MJjainam, line 354: Should we be not printing both?

action.yml
MJjainam: Can the output and error be interleaved that we might be missing in sequential cat commands at line 302 to 35

action.yml
```diff
@@ -289,16 +289,19 @@ runs:
...
-          fi
+        fi
+
+        # We capture stdout (JSON) to TEMP_STDOUT and stderr to TEMP_STDERR
+        if ! gemini --yolo --prompt "${PROMPT}" --output-format json 2> "${TEMP_STDERR}" 1> "${TEMP_STDOUT}"; then
+          FAILED=true
+        fi
+
+        if [[ "${DEBUG}" = true ]]; then
+          cat "${TEMP_STDOUT}"
```
MJjainam, line 304: We must make it explicit that this is stdout and stderr.

action.yml
```diff
@@ -312,20 +316,48 @@ runs:
...
+        RESPONSE=""
+        ERROR_JSON=""
+        if jq -e . "${TEMP_STDOUT}" >/dev/null 2>&1; then
+           RESPONSE=$(jq -r '.response // ""' "${TEMP_STDOUT}")
+        fi
+        if jq -e . "${TEMP_STDERR}" >/dev/null 2>&1; then
+           ERROR_JSON=$(jq -c '.error // empty' "${TEMP_STDERR}")
+        fi
+        if [[ -z "${RESPONSE}" && -z "${ERROR_JSON}" ]]; then
+           echo "::warning::Gemini CLI output was not valid JSON"
```
MJjainam: Can response and error be empty but Gemini CLI output still be a valid json? We must probably change the logic to a simpler if jq -e . {} else {echo gemini cli output was not valid json}

action.yml
```diff
@@ -312,20 +316,48 @@ runs:
...
+        fi
+
+
         # Set the captured response as a step output, supporting multiline
         echo "gemini_response<<EOF" >> "${GITHUB_OUTPUT}"
-        cat "${TEMP_STDOUT}" >> "${GITHUB_OUTPUT}"
+        if [[ -n "${RESPONSE}" ]]; then
+          echo "${RESPONSE}" >> "${GITHUB_OUTPUT}"
+        else
+          cat "${TEMP_STDOUT}" >> "${GITHUB_OUTPUT}"
```
MJjainam, line 339: This would be a change in behavior from before. 

TEMP_STDOUT would contain extra metadata which was not present before. 
Is this intended?

action.yml
```diff
@@ -312,20 +316,48 @@ runs:
...
+        fi  
         echo "EOF" >> "${GITHUB_OUTPUT}"
 
         # Set the captured errors as a step output, supporting multiline
         echo "gemini_errors<<EOF" >> "${GITHUB_OUTPUT}"
-        cat "${TEMP_STDERR}" >> "${GITHUB_OUTPUT}"
+        if [[ -n "${ERROR_JSON}" ]]; then
+          echo "${ERROR_JSON}" >> "${GITHUB_OUTPUT}"
```
MJjainam, line 346: There is a tricky nuance in this

ERROR_JSON is the error coming back from Gemini CLI internals (think LLM)
TEMP_STDERR is the error coming from the outer Gemini CLI

I think this changes behavior. 

@megha1188 could you validate and add your suggestion what best can be done?

action.yml
gsehgal: The new logic is based on the following rationale:
if TEMP_STDOUT had valid JSON then RESPONSE will have value which we print in line 337.
if TEMP_STDOUT did not have valid JSON then we print that to GITHUB_OUTPUT in line 339.

action.yml
gsehgal: No Response and error will not be empty ever. It always will give you some result back, CLI is an agent, when you give it a task it will definitely respond with some value.

action.yml
gsehgal: As per my observation, and also inspecting the code again; in case of JSON output everything is being captured in JSON.
https://github.com/google-gemini/gemini-cli/blob/main/packages/cli/src/nonInteractiveCli.ts#L338

action.yml
MJjainam: From code readability perspective, it would benefit to have an explicit check for invalid json. 
Although the above works, the reader might be confused on the specific IF scenario where invalid json statement is printed.

action.yml
megha1188: I see that the whole error message is getting printed here, So specific error code handling is not needed here unless we do a specific action on that error code. Currently, since there is no specific action on error code, this looks good.

action.yml
gsehgal: printing TEMP_STDERR by default.

gsehgal (commented):
Addressed comments.

action.yml
MJjainam: The strings Start and End Gemini CLI STDERR will be printed even when TMP_STDERR is empty right?

gsehgal: if Failed==true , then TEMP_STDERR will have content.

scidomino (approved):

PR #405, opened 2025-12-03
Update the gemini review toml to send correct response with no tags

megha1188:
This PR updates the gemini-review.toml prompt template to remove several extra newline characters.

These newlines were causing the model to sometimes generate responses with incorrect formatting or unwanted tags. Removing them cleans up the prompt and ensures the model's output more reliably adheres to the desired structure.

The change is applied to both the default command prompt and the example workflow prompt.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "run-gemini-cli",
-  "version": "0.1.15",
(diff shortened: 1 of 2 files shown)


megha1188:
@gemini-cli /review

MJjainam:
@gemini-cli /review

PR #407, opened 2025-12-05
Rename `DEBUG` input to `GEMINI_DEBUG` and pass `--debug` flag to CLI.

parthasaradhie:
related to https://github.com/google-github-actions/run-gemini-cli/issues/337

Env variable DEBUG is used in [CONTRIBUTING.md?plain=1#L356 ](https://github.com/google-gemini/gemini-cli/blob/934b309b4cc66925d59ec671b1a8eb2bd44ef8ef/CONTRIBUTING.md?plain=1#L356 ) 
Setting this to true will start gemini in inspect-brk mode which will wait for node debugger to attach indefinitely. resulting in looks like freeze. 

so i have removed the support of the env variable and added warning to users to notify it side effects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,7 +213,7 @@ We recommend setting the following values as repository variables so they can be
 
 | Name                        | Description                                                                                                                    | Type     | Required | When Required                  |
 | --------------------------- | ------------------------------------------------------------------------------------------------------------------------------ | -------- | -------- | ------------------------------ |
-| `DEBUG`                     | Enables debug logging for the Gemini CLI.                                                                                      | Variable | No       | Never                          |
+| `GEMINI_DEBUG`              | Enables debug logging for the Gemini CLI.                                                                                      | Variable | No       | Never                          |
 | `GEMINI_CLI_VERSION`        | Controls which version of the Gemini CLI is installed.                                                                         | Variable | No       | Pinning the CLI version        |
 | `GCP_WIF_PROVIDER`          | Full resource name of the Workload Identity Provider.                                                                          | Variable | No       | Using Google Cloud             |
 | `GOOGLE_CLOUD_PROJECT`      | Google Cloud project for inference and observability.                                                                          | Variable | No       | Using Google Cloud             |
--- a/.github/workflows/gemini-dispatch.yml
+++ b/.github/workflows/gemini-dispatch.yml
@@ -25,7 +25,7 @@ defaults:
 jobs:
   debugger:
     if: |-
-      ${{ fromJSON(vars.DEBUG || vars.ACTIONS_STEP_DEBUG || false) }}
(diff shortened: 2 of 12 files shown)


parthasaradhie:
@gemini-cli /review

MJjainam:
@gemini-cli /review

action.yml
```diff
@@ -290,19 +290,20 @@ runs:
...
-          echo "::warning::Gemini CLI debug logging is enabled. This will stream responses, which could reveal sensitive information if processed with untrusted inputs."
-        fi
-
         # We capture stdout (JSON) to TEMP_STDOUT and stderr to TEMP_STDERR
-        if ! gemini --yolo --prompt "${PROMPT}" --output-format json 2> "${TEMP_STDERR}" 1> "${TEMP_STDOUT}"; then
-          FAILED=true
-        fi
-
-        if [[ "${DEBUG}" = true ]]; then
+        if [[ "${GEMINI_DEBUG}" = true ]]; then
```
MJjainam, line 294: Note: This would be a breaking change for many existing workflows that use the variable "DEBUG". 

Also, DEBUG=1 is only applicable for sandbox mode - should we instead change that variable?
This would also benefit the confusion regarding --debug vs DEBUG=true env variable.

parthasaradhie: This repo is still in Beta state and env variables are expected to change if its breaking the workflows. 
The main reason "DEBUG" should not be supported is due to the upstream gemini cli repo uses this internally for actual debug (breakpoint and attaching to node process) related tasks and is explicitly mentioned in gemini-cli repo [doc](https://github.com/google-gemini/gemini-cli/blob/7a720375729cc1e5b92ec02910a2fa8be77aa05d/docs/cli/sandbox.md?plain=1#L143)   
i ack that --debug needs to use, i will update the flag in the code. (even i thought it, but before adding it want to know from maintainers perspective why they left it out before).

README.md
```diff
@@ -213,7 +213,6 @@ We recommend setting the following values as repository variables so they can be
 
 | Name                        | Description                                            | Type     | Required | When Required             |
 | --------------------------- | ------------------------------------------------------ | -------- | -------- | ------------------------- |
```
MJjainam: GEMINI_DEBUG needs to be added.

parthasaradhie: This is getting controlled via gemini_debug var above, no need to be set this as repo variable

MJjainam:
Should we update all the vars.DEBUG variable to vars.GEMINI_DEBUG ?

parthasaradhie:
> Should we update all the vars.DEBUG variable to vars.GEMINI_DEBUG ?

Updated all the references

README.md
MJjainam: GEMINI_DEBUG is added.

Issue #373, opened 2025-10-23
Refactor logic out of YAML/Bash and into a TypeScript package


jerop:
**Description:**

The `action.yml` currently contains significant amounts of complex logic embedded directly within `bash` `run` steps. This approach makes the action difficult to test, maintain, and debug. Refactoring this logic into a dedicated TypeScript  package will improve code quality, enable proper unit testing, and enhance reusability.

**Specific areas for refactoring:**

1.  **Input Validation Logic:** The `Validate Inputs` step contains extensive conditional logic for validating authentication methods and input combinations. This should be moved into a TypeScript module.
2.  **Gemini CLI Installation Logic:** The `Install Gemini CLI` step handles different installation methods based on the `gemini_cli_version`. This logic, including npm and git operations, should be encapsulated in a TypeScript function.
3.  **Gemini CLI Execution and Output Handling:** The `Run Gemini CLI` step manages the execution of the Gemini CLI, captures its stdout/stderr, and handles error conditions. The output parsing and error reporting could be made more robust and testable in a TypeScript module.
4.  **Telemetry Upload Logic:** The `Upload Telemetry to Google Cloud` step involves Docker commands, file manipulation, and a polling mechanism. This complex sequence should be refactored into a dedicated TypeScript module for better testability and maintainability.

**Acceptance Criteria:**

*   All identified complex bash logic is replaced with calls to functions within a new TypeScript package.
*   The `action.yml` remains focused on orchestrating steps rather than implementing detailed logic.
*   A clear project structure for the TypeScript package is established (e.g., `src/`, `dist/`, `package.json`).

jerop:
@gemini-cli /triage

PR #409, opened 2025-12-08
chore: Add Prettier and pre-commit hook for consistent formatting, applying initial format fixes across documentation and examples.

parthasaradhie:
Adding prettier config as default formatter. and pre commit hook to format files automatically.
Reformatting existing file
This is part of https://github.com/google-github-actions/run-gemini-cli/issues/373

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,9 @@
+npx lint-staged || {
+  echo ''
+  echo '===================================================='
+  echo 'pre-commit checks failed. in case of emergency, run:'
+  echo ''
+  echo 'git commit --no-verify'
+  echo '===================================================='
+  exit 1
+}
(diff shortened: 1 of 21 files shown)


MJjainam:
@gemini-cli /review

.husky/pre-commit
```diff
@@ -0,0 +1,9 @@
+npx lint-staged || {
```
MJjainam, line 1: What if the lint-staged is not available?

.husky/pre-commit
MJjainam, line 1: @gemini-cli review this file

.husky/pre-commit
parthasaradhie: lint-staged will be available when dev does ```npm install``` its added as dev dependency

Issue #375, opened 2025-10-23
Add unit tests for critical logic in Typescript package


jerop:
**Description:**

Currently, there are no apparent automated tests for the `run-gemini-cli` action. To ensure the reliability and correctness of the action, comprehensive test coverage is required, especially for the critical logic branches identified during refactoring.

Depends on https://github.com/google-github-actions/run-gemini-cli/issues/373

**Specific testing requirements:**

1.  **Unit Tests for TypeScript Modules:**
    *   Develop unit tests for all functions and modules created during the refactoring process (e.g., input validation, installation logic, telemetry upload).
    *   Achieve high code coverage for these modules, focusing on all conditional branches and edge cases.
2.  **Mocking External Dependencies:**
    *   Mock external calls (e.g., `npm`, `git`, `docker`, `curl`, `gemini CLI`) to ensure unit tests are fast and isolated.

**Acceptance Criteria:**
*   A testing framework is integrated into the project.
*   Unit tests are written for all refactored TypeScript modules.
*   Test coverage reports demonstrate significant coverage of the action's core logic.
*   Tests can be run easily via a `npm` script.

jerop:
@gemini-cli /triage

PR #422, opened 2025-12-10
Moving validate logic to Typescript

parthasaradhie:
### Refactor Input Validation Logic
The Validate Inputs step contains extensive conditional logic for validating authentication methods and input combinations. This is moved into a TypeScript module.

### Related Issues

https://github.com/google-github-actions/run-gemini-cli/issues/373 
https://github.com/google-github-actions/run-gemini-cli/issues/375

### Verification
manually verified the new logic by triggering an related action in personal  github [repo](https://github.com/parthasaradhie/hello-world-rpm/blob/f940b18353fa2ef6904bc92df0b8909e0f8b5fdd/.github/workflows/release-with-gemini.yaml#L95) 
i set both gemini_api_key and google_api_key which should trigger warning about using both auth tokens at once and it displayed as expected. 

run link (expected to fail due to set config): https://github.com/parthasaradhie/hello-world-rpm/actions/runs/20143676300/job/57818088090
warning: 
image

workflow link: [release-with-gemini.yaml#L95](https://github.com/parthasaradhie/hello-world-rpm/blob/f940b18353fa2ef6904bc92df0b8909e0f8b5fdd/.github/workflows/release-with-gemini.yaml#L95) 

workflow config: 
image


src/validate-inputs.ts
```diff
@@ -0,0 +1,96 @@
...
+      useVertexAi,
+      useGeminiCodeAssist,
+    } = geminiCliActionsInputs();
+
+    const warn = (msg: string) => {
+      core.warning(msg, { title: 'Input validation' });
+      core.summary
+        .addHeading('Input validation warnings', 3)
+        .addList([msg])
+        .write();
```
galdawave, line 28: The core.summary.write() method returns a Promise<void>. In the warn function, this promise is ignored. If the script finishes execution before the asynchronous write completes, the summary might not be written to the GITHUB_STEP_SUMMARY file and calling .write() multiple times (once per warning) triggers multiple file system operations. Refactor to accumulate all warnings in a list and write to the summary once at the end of execution, ensuring to await the write operation. You will need to make validateInputs async.

action.yml
```diff
@@ -103,77 +103,34 @@ outputs:
 runs:
   using: 'composite'
   steps:
-    - name: 'Validate Inputs'
-      id: 'validate_inputs'
+    - name: 'Install pnpm'
```
galdawave, line 106: This adds network and I/O overhead to every workflow using this action, which was previously a near-instantaneous shell script. Consider bundling the TypeScript code into a single JavaScript file and committing it to the repository (e.g., dist/validate_inputs.js). This would allow you to run node dist/validate_inputs.js directly without needing to install dependencies at runtime.

parthasaradhie: You're right— The ```dist/*.js``` approach is good and migration to this is planned soon. Once #373 is resolved, I’ll open a new issue to migrate the [composite action in action.yml ](https://github.com/google-github-actions/run-gemini-cli/blob/87c4941afde98641c57b70f672fbdfb7b9fa4623/action.yml#L104) to a Node.js action (using: 'node20'). This will have the logic to run the js file directly from dist folder and it also streamline input parsing and improve control. I’m targeting this for the v0.18 release. In the meantime, the overhead is minimal; a clean install takes only about 2–3 seconds locally and might take less in CI environments.
```
computer-name$ npm cache clean -f
npm warn using --force Recommended protections disabled.
computer-name$ rm -rf node_modules/
computer-name$ npm i

> run-gemini-cli@0.1.16 prepare
> husky


added 134 packages, and audited 135 packages in 2s

40 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
```

src/validate-inputs.ts
parthasaradhie: > the summary might not be written to the GITHUB_STEP_SUMMARY file 

Node js event loop keeps the process alive as long as there are pending asynchronous tasks, so this wont miss.

> calling .write() multiple times (once per warning) triggers multiple file system operations

This is intentional for the current scope, 

- There are total of 8 lines of file operations happening here and for now and there there won't be any significance performance gain/lose can happen here.
- Prioritizing code simplicity and maintainability for now. While collecting logs into a list is a valid optimization, the current approach serves as a reasonable trade-off until we streamline our overall logging and input validation flow during the upcoming TypeScript migration.

parthasaradhie:
@gemini-cli /review

parthasaradhie:
Above comment is for testing new release

parthasaradhie:
@gemini-cli /review

Issue #399, opened 2025-11-25
Custom commands overwritten by action


dan-ri:
The [instructions](https://github.com/google-github-actions/run-gemini-cli/tree/main/examples/workflows/pr-review#review-prompt-customization) for custom commands don't work because the action unconditionally [overwrites them](https://github.com/google-github-actions/run-gemini-cli/blob/main/action.yml#L200-L207).

There are also documentation inconsistencies, such as [instructing](https://github.com/google-github-actions/run-gemini-cli/tree/main/examples/workflows/pr-review#prerequisites) the user to add `.gemini` to `.gitignore` despite also instructing them to place custom commands in this directory.

GrinZero:
It's same!!!

jerop:
@gemini-cli /triage

PR #432, opened 2025-12-24
fix: install custom commands from workspace or current action

GrinZero:
Fix: https://github.com/google-github-actions/run-gemini-cli/issues/399

I added the workspace path so that GitHub Actions will first install the commands from this repository by default, and then, if it detects that the user’s repository contains custom commands, it will override them with the commands from the user’s repository.


GrinZero:
@MJjainam I noticed that you've been the only one active recently. If you have time, please help review it

PR #480, opened 2026-03-11
fix: Add file-based output mode to avoid exceeding GitHub Actions output limits

### LINKED ISSUE #479: Add file-based output mode to avoid exceeding GitHub Actions output limits with debug enabled
### TL;DR

When `gemini_debug` is enabled, the action streams all debug output to the console via `tee` and writes the full content to `GITHUB_OUTPUT`. Recent versions of Gemini CLI (post-0.25.x) significantly increased the volume of diagnostic output behind the `--debug` flag (auth info, API request/error counts, retry events, loop detection). This causes `GITHUB_OUTPUT` to exceed the ~1MB platform limit, resulting in truncated or failed step outputs.

### Proposal

Add a new `output_to_file` boolean input (default: `false`) that, when enabled:

1. **Redirects all output to files only** — runs with `--debug` but without `tee` to the console, same as non-debug redirection but with the `--debug` flag
2. **Returns file paths as step outputs** — `summary` contains the path to `gemini-artifacts/stdout.log` and `error` contains the path to `gemini-artifacts/stderr.log`, instead of the file contents
3. **Produces a compact job summary** — instead of dumping the full response/error into the step summary, shows file sizes and paths
4. **Adds new outputs** — `output_mode` (`content` or `file`) and `artifacts_dir` (path to `gemini-artifacts/`) so callers can programmatically detect the mode

When `output_to_file` is not set (or `false`), behavior is **identical to today** — full backwards compatibility.

### Why file-based output?

- The `gemini-artifacts/` directory already exists and is populated with `stdout.log`, `stderr.log`, and `telemetry.log` on every run
- Callers can upload these as artifacts (via `upload_artifacts: true`), filter/process them in subsequent steps, or extract specific data (e.g., stats, traces)
- This pattern naturally enables future `stream-json` adoption — once output goes to a file, callers can parse NDJSON to extract both the final response and a step-by-step agent trace

### Detailed design

The execution block (lines 294-307) would add a third branch:

```bash
if [[ "${GEMINI_DEBUG}" = true ]] && [[ "${OUTPUT_TO_FILE}" = true ]]; then
  # Debug to files only — no console streaming
  if ! gemini --debug --yolo --prompt "${PROMPT}" --output-format json \
    2> "${TEMP_STDERR}" 1> "${TEMP_STDOUT}"; then
    FAILED=true
  fi
elif [[ "${GEMINI_DEBUG}" = true ]]; then
  # Existing behavior: stream to console via tee
  ...
else
  # Existing behavior: silent capture
  ...
fi
```

The output-setting block (lines 340-356) would conditionally emit paths or content:

```bash
if [[ "${OUTPUT_TO_FILE}" = true ]]; then
  echo "gemini_response<<EOF" >> "${GITHUB_OUTPUT}"
  echo "$(pwd)/gemini-artifacts/stdout.log" >> "${GITHUB_OUTPUT}"
else
  # existing content-based output
fi
```

### Related issues

- #381 — `summary` output contains invalid JSON due to `EOF` delimiter handling. File-based output sidesteps this entirely since file paths don't contain `EOF`.
- #305 — Refactor scripts into own files. This change adds a small amount of logic to the existing bash block; if the refactor lands first, the change would go into the extracted script instead.

### Additional information

I'm happy to submit a PR for this. My use case is a CI code review bot that uses `gemini_debug: true` with 80+ turn sessions, producing debug output that consistently exceeds GitHub's output limits.

michalszelagsonos:
## Summary

- Fixes #479
- Add `output_to_file` boolean input that redirects all output to files and returns file paths as step outputs instead of content
- Works independently of `gemini_debug` — when enabled, output always goes to files regardless of debug mode
- New outputs: `output_mode` (content/file) and `artifacts_dir` (path to gemini-artifacts/)

## Motivation

When `gemini_debug` is enabled, recent gemini-cli versions (post-0.25.x) significantly increased the volume of diagnostic output behind `--debug` (auth info, API request/error counts, retry events, loop detection). This causes `GITHUB_OUTPUT` to exceed the ~1MB platform limit, resulting in truncated or failed step outputs.

The `gemini-artifacts/` directory already exists and is populated on every run — this change leverages it by returning file paths instead of dumping content into step outputs.

## Changes

**New input** (`output_to_file`, default `false`):
- When `true`, redirects stdout/stderr to files (no `tee`/console streaming)
- Returns file paths (`gemini-artifacts/stdout.log`, `gemini-artifacts/stderr.log`) as `summary` and `error` outputs instead of content
- Produces a compact job summary showing file sizes instead of full content
- Works with or without `gemini_debug` — the debug flag only controls whether `--debug` is passed to the CLI

**New outputs**:
- `output_mode`: `"content"` (default) or `"file"` — lets callers detect the mode programmatically
- `artifacts_dir`: absolute path to `gemini-artifacts/` directory

**Refactored execution block**:
- Extracted `GEMINI_ARGS` to avoid duplicating the gemini CLI invocation across branches
- `output_to_file` is checked first, independently of `gemini_debug`
- The existing `tee`-based streaming only applies when `output_to_file` is false and `gemini_debug` is true

**Backwards compatibility**: When `output_to_file` is not set or `false`, behavior is identical to today.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,9 +190,15 @@ go to the [Gemini Assistant workflow documentation](./examples/workflows/gemini-
 
 -   <a name="__input_upload_artifacts"></a><a href="#user-content-__input_upload_artifacts"><code>upload_artifacts</code></a>: _(Optional, default: `false`)_ Whether to upload artifacts to the github action.
 
+-   <a name="__input_output_to_file"></a><a href="#user-content-__input_output_to_file"><code>output_to_file</code></a>: _(Optional, default: `false`)_ When true, redirect all output to files and return file paths as step outputs instead of content. Useful when gemini_debug is true to avoid exceeding GitHub Actions output limits.
+
 -   <a name="__input_use_pnpm"></a><a href="#user-content-__input_use_pnpm"><code>use_pnpm</code></a>: _(Optional, default: `false`)_ Whether or not to use pnpm instead of npm to install gemini-cli
 
 -   <a name="__input_workflow_name"></a><a href="#user-content-__input_workflow_name"><code>workflow_name</code></a>: _(Optional, default: `${{ github.workflow }}`)_ The GitHub workflow name, used for telemetry purposes.
+
+-   <a name="__input_github_pr_number"></a><a href="#user-content-__input_github_pr_number"><code>github_pr_number</code></a>: _(Optional, default: `${{ github.event.pull_request.number }}`)_ The Pull Request number the CLI is operating on. Defaults to the event payload.
+
+-   <a name="__input_github_issue_number"></a><a href="#user-content-__input_github_issue_number"><code>github_issue_number</code></a>: _(Optional, default: `${{ github.event.issue.number }}`)_ The Issue number (or comma-separated list of issue numbers) the CLI is operating on. Defaults to the event payload.
 
 
 <!-- END_AUTOGEN_INPUTS -->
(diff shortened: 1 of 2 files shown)


dashesy:
Thanks for working on this. One concern for our use case: as written, `output_to_file=true` uses `2> "${TEMP_STDERR}" 1> "${TEMP_STDOUT}"` (no `tee`), which removes the live streaming to the step log that `gemini_debug=true` currently enables.

Our need is specifically to watch long-running PR reviews live (sometimes 20+ minutes) to detect stuck states — artifacts available only post-run don't help there.

Would it be possible to keep the `tee` streaming in debug mode, and only change the fallback that writes raw `TEMP_STDOUT` to `$GITHUB_OUTPUT`? e.g. emit a placeholder:

```bash
echo "(debug output — see artifacts)" >> "${GITHUB_OUTPUT}"
```

That would preserve live streaming while avoiding the template-engine size crash.

michalszelagsonos:
Yeah, I think I can put that live stream back in.  I can see how that is useful for live monitoring or to just have a trace on long running tasks.

PR #505, opened 2026-04-24
Modernize example workflows for trust and security

ehedlund:
This PR updates the example workflows and documentation to follow the latest trust and security guidelines.

Key changes:
- Added `GEMINI_TRUST_WORKSPACE: 'true'` to all example workflows.
- Replaced unsafe shell tools with native tools (`list_directory`, `read_file`, `grep_search`).
- Updated triage workflows to return results via the action's `summary` output instead of using `echo` to set environment variables.
- Explicitly added `sandbox: false` to the tools configuration.
- Corrected technical inaccuracies in `docs/trust-guidance.md`.


examples/workflows/gemini-assistant/gemini-invoke.yml
```diff
@@ -111,12 +112,12 @@ jobs:
...
                 "core": [
-                  "run_shell_command(cat)",
-                  "run_shell_command(echo)",
-                  "run_shell_command(grep)",
-                  "run_shell_command(head)",
-                  "run_shell_command(tail)"
-                ]
+                  "list_directory",
+                  "read_file",
+                  "grep_search",
```
amlweems: Are these tools needed given the mcp above? (with get_commit and get_file_contents)

examples/workflows/gemini-assistant/gemini-invoke.yml
```diff
@@ -111,12 +112,12 @@ jobs:
...
+                  "run_shell_command(echo)"
```
amlweems: This comment applies to any workflow that allowlists echo: 

iiuc this is required due to the way we pass inputs in the prompt (e.g. `!{echo $TITLE}`). Is it feasible to interpolate these env vars directly into the prompt / or pass them some other way? If we could eliminate run_shell_command I think that would also improve harness reliability since it wouldn't be trying to call run_shell_command ls, cat, grep etc (since the tool just wouldn't exist in the system prompt).

examples/workflows/gemini-assistant/gemini-invoke.yml
amlweems, line 41: Could use test turning off persist credentials to this (and other workflows that use checkout)?
```
        with:
          persist-credentials: 'false'
```
By default, GitHub writes the GH token to .git/config, which is in the local directory when Gemini CLI runs, so with folder trust, it can read this file.

kschaab (approved):

examples/workflows/CONFIGURATION.md
```diff
@@ -101,9 +101,9 @@ with:
     {
       "tools": {
         "core": [
+          "list_directory",
           "read_file",
-          "run_shell_command(echo)",
```
kschaab, line 105: I don't know if you want to keep this to continue the example of using `run_shell_command` in the configuration with echo?

kschaab (approved):

examples/workflows/CONFIGURATION.md
ghost: Good point thanks!

examples/workflows/gemini-assistant/gemini-invoke.yml
ghost: Good point, we shouldn't need these - removed

examples/workflows/gemini-assistant/gemini-invoke.yml
ghost: Updated!

examples/workflows/gemini-assistant/gemini-invoke.yml
ghost: Instead of interpolating via shell `echo`, the workflows now include a setup step that uses `jq` to safely package all necessary environment variables into a single `.gemini/context.json` file. The `.toml` prompts have been updated to load this file directly using the native `!{read_file('.gemini/context.json')}` tool. Let me know if this is ok!

amlweems: Wow, this is great! The only question I have is how read_file works with the new WORKSPACE_TRUST. can Gemini only read files in the local directory, or can it read any file?

ghost: It can only read files in the workspace!

amlweems (commented):
LGTM!

